PR #5, opened 2023-06-12
Add i18n & router

HyTekCoop:
Add i18n & router and also delete empty files


src/application/router/router.ts
neSpecc, line 1: lets call the file `index.ts`

neSpecc (commented):
seems like i18 and router are not connected to the `src/index.ts`

src/application/i18n/index.ts
```diff
@@ -1,5 +1,15 @@
 import Vue from 'vue';
 import VueI18n from 'vue-i18n';
-import languages from './languages/en'
+import en from './messages/en.json';
 
-Vue.use(VueI18n);
\ No newline at end of file
+Vue.use(VueI18n);
```
slaveeks, line 5: I think, you need to call it in `index.ts` file

neSpecc (changes requested):

src/application/router/index.ts
```diff
@@ -0,0 +1,14 @@
+import { createRouter, createWebHistory } from 'vue-router';
+
+const router = createRouter({
+  history: createWebHistory(),
+  routes: [
+    {
+      path: '/',
+      redirect: '/app',
```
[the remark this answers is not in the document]
HyTekCoop: changes added

neSpecc (approved):

slaveeks (approved):

PR #6, opened 2023-06-14
fix(build): build errors fixed

neSpecc:
Currently the build in the `main` branch is broken after #2 

This PR fixes build error and removes redundant files.


src/application/i18n/index.ts
ilyamore88, line 1: Why didn't you just delete this file?

neSpecc: it will be used later

ilyamore88 (approved):

n0str (approved):

PR #7, opened 2023-06-14
Docker support

n0str:


ilyamore88 (approved):

vite.config.ts
```diff
@@ -2,4 +2,7 @@ import { defineConfig } from 'vite';
 import vue from '@vitejs/plugin-vue';
 export default defineConfig({
   plugins: [ vue() ],
+  server: {
+    host: true
+  }
```
neSpecc, line 7: is it necessary?

n0str: not necessary

neSpecc (approved):

PR #8, opened 2023-06-18
Layouts: Add default layout

kseniaSs:
Added default layout 

After added vue-router, we have to add $route.meta.layout to AppLayout

Right now, it's just default component

P.S. I didn't add a lot of styles because i'm not sure about responsibility of Layouts.


ilyamore88 (approved):

src/App.vue
```diff
@@ -1,6 +1,23 @@
...
+});
+
+provide("app:layout", layout);
+
+</script>
+
 <template>
   <div id="app">
-    <router-view></router-view>
+    <component :is="layout || 'div'">
```
neSpecc, line 18: maybe it would be cleaner to move this part to the `presentation/layouts/Layout.vue` and get rid of `layouts/index.ts`?

neSpecc, line 18: lets rename AppLayout to the Default.vue and use it by default instead of 'div'

src/presentation/layouts/AppLayout.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <div :is="layout" class="center-container">
+    <slot />
+  </div>
+</template>
+
+<script>
```
neSpecc, line 7:
```suggestion
<script lang="ts"> 
```

src/application/router/routes.ts
```diff
@@ -4,6 +4,9 @@ import Note from '@/presentation/pages/Note.vue';
 const routes = [
   {
     path: '/',
+    meta: {
+      layout: 'DefaultLayout',
```
neSpecc, line 8: could this property be optional so we won't include it if the Default layout is used

src/application/services/useLayout.ts
```diff
@@ -0,0 +1,23 @@
...
+ * useLayout hook is setting current Layout
+ *
+ * @returns { ShallowRef<string> } - Layout template was got by meta
+ */
+export default function (): ShallowRef<string> {
+  const layout = shallowRef('div');
+
+  router.afterEach((to) => {
+    layout.value = layouts[to.meta.layout] || 'div';
+  });
```
neSpecc, line 19: I think, you should use the current route instead of looping over all routes

src/presentation/layouts/DefaultLayout.vue
neSpecc, line 1: lets rename the file to `Default.vue` since it already under `/layouts/` directory

src/presentation/layouts/Layout.vue
```diff
@@ -0,0 +1,17 @@
+<template>
+  <component :is="layout || DefaultLayout">
```
neSpecc, line 2: this logic could be placed at the `useLayout` composable

src/vite-env.d.ts
```diff
@@ -1,2 +1,9 @@
 // / <reference types="vite/client" />
-declare module '*.vue'
\ No newline at end of file
+declare module '*.vue'
+
+declare module 'vue-router' {
+    interface RouteMeta {
+      // add Layout like component wrapper
```
neSpecc, line 6: please, use block-style comments

src/application/services/useLayout.ts
kseniaSs: Why is is loop? It's just once after rendering, because lProperty "layout" was accessed during render but is not defined on instance.

src/presentation/layouts/Default.vue
```diff
@@ -0,0 +1,18 @@
+<template>
+  <div :is="layout" class="center-container">
+    <slot />
+  </div>
+</template>
+
+<script lang="ts">
```
neSpecc, line 7: use script setup pls

src/App.vue
```diff
@@ -1,10 +1,14 @@
...
   <Header />
-  <router-view></router-view>
+  <Layout>
+    <router-view />
+  </Layout>
 </template>
 
 <script lang="ts" setup>
 import Header from '@/presentation/components/Header.vue';
+import useLayout from '@/application/services/useLayout';
```
neSpecc, line 10: У нас eslint не подсвечивает неиспользуемые импорты?
```suggestion
```

kseniaSs: nope :c

neSpecc (approved):

PR #14, opened 2023-07-07
Added aliases

slaveeks:
Added aliases for:
1. `/src` - `@/`
2. `/src/application` -> `@application/`
3. `/src/presentation` -> `@presentation/`
4. `/src/infrastructure` -> `@infrastructure/`
5. `/src/domain` -> `@domain/`

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -1,6 +1,6 @@
 import { Ref, ref } from 'vue';
-import Note from '@domain/entities/Note';
-import { noteService } from '@domain/index';
+import Note from '@/domain/entities/Note';
+import { noteService } from '@/domain/index';
 
 /**
  * Note hook state
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -1,6 +1,6 @@
 import { Ref, ref } from 'vue';
 import Note from '@/domain/entities/Note';
-import { noteService } from '@/domain/index';
+import { noteService } from '@/domain';
 
 /**
  * Note hook state
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -1,6 +1,6 @@
 import { Ref, ref } from 'vue';
-import Note from '../../domain/entities/Note';
-import { noteService } from '../../domain';
+import Note from '@domain/entities/Note';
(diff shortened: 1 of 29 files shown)


neSpecc:
More common way is just add '@' as alias to 'src', so urls will look better:

```
@/application/...
@/domain/...
```

For example, Nuxt has such aliases

tsconfig.json
```diff
@@ -12,6 +13,12 @@
     "noEmit": true,
     "jsx": "preserve",
     "allowSyntheticDefaultImports": true,
+    "paths": {
+      "@/infrastructure/*": ["infrastructure/*"],
+      "@/presentation/*": ["presentation/*"],
+      "@/domain/*": ["domain/*"],
+      "@/application/*": ["application/*"]
```
neSpecc, line 20: did you tried this?
```suggestion
      "@/*": ["./src/*"],
```

slaveeks: it is not working in my IDE

slaveeks: but i can simplify vite.config

neSpecc: Try changing the `include` property:

```
"./src/**/*.ts",
"./src/**/*.vue",
```

and vite alias:
```
'@': path.resolve(__dirname, './src'),
```

slaveeks: `'@/': '/src/',`, i can make this in vite alias
Changing `include` property does not help

src/application/services/useNote.ts
```diff
@@ -1,6 +1,6 @@
 import { Ref, ref } from 'vue';
-import Note from '../../domain/entities/Note';
-import { noteService } from '../../domain';
+import Note from '@/domain/entities/Note';
+import { noteService } from '@/domain/index';
```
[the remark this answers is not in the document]
slaveeks: fixed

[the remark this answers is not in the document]
kseniaSs: but `import { noteService } from '@/domain/index';` still exists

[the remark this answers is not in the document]
slaveeks: fixed :)

kseniaSs (approved):

neSpecc (approved):

PR #16, opened 2023-07-12
Themes: add light | dark theme

kseniaSs:
Added hook for changing theme. Added background variables as examples of using theme.

### DIFF
--- a/src/App.vue
+++ b/src/App.vue
@@ -15,12 +15,12 @@ export default {
 </script>
 
 <style scoped>
-  * {
-    box-sizing: border-box;
+  html,
+  body {
+    height: 100%;
   }
 
   #app {
-    background-color: var(--background);
     min-height: 100%;
     background: var(--color-bg);
     color: var(--color-text-main);
--- a/src/App.vue
+++ b/src/App.vue
@@ -7,9 +7,7 @@
 <script lang="ts">
 import { useColorMode } from '@vueuse/core'
 
-const mode = useColorMode({
-  initialValue: "dark"
-})
+const mode = useColorMode()
 
 export default {
   name: "App",
@@ -17,8 +15,6 @@ export default {
 </script>
 
 <style scoped>
-  @import "./presentation/styles/colors.pcss";
-
   #app {
     background-color: var(--background);
   }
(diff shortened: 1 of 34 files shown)


src/App.vue
```diff
@@ -5,11 +5,22 @@
 </template>
 
 <script lang="ts">
+import { useColorMode } from '@vueuse/core'
+
+const mode = useColorMode({
+  emitAuto: true
```
neSpecc, line 11: Looks like `emitAuto` is deprecated: https://github.com/vueuse/vueuse/pull/2980/files#diff-5d628cd173c95cd5c9a1a70e3299bf4d58cceb8bcc22a31baf0cf7793f9db740R82

src/presentation/styles/colors.pcss
```diff
@@ -0,0 +1,7 @@
+:global(.light) {
+  --background: #fff;
+}
```
neSpecc, line 3: When document won't have a `.light` or `.dark` class, there will be no variable values specified? Let's make `dark` values be specified by default

src/presentation/styles/global.pcss
```diff
@@ -0,0 +1,3 @@
+:global(body) {
```
neSpecc, line 1: maybe we should use [normalize.css](https://necolas.github.io/normalize.css/) instead?

src/App.vue
```diff
@@ -5,11 +5,21 @@
 </template>
 
 <script lang="ts">
+import { useColorMode } from '@vueuse/core'
+
+const mode = useColorMode({
+  initialValue: "dark"
```
slaveeks, line 11: Может сразу `auto` поставим?

src/App.vue
```diff
@@ -5,11 +5,21 @@
...
+})
+
 export default {
   name: "App",
 }
 </script>
 
 <style scoped>
+  @import "./presentation/styles/colors.pcss";
```
slaveeks, line 20: Can we import it in `index.ts` to omit it in components?

neSpecc:
Please, update the branch from `main`. Color variables were added

kseniaSs:
> Please, update the branch from `main`. Color variables were added

did it

PR #20, opened 2023-07-20
Added Source Sans Pro font

slaveeks:
1. Added `assets/fonts/SourceSansPro`
2. Added `fonts.pcss`
3. Set font-family to `Source Sans Pro`

image


src/App.vue
```diff
@@ -12,5 +12,6 @@ import Header from './presentation/components/Header.vue';
 <style>
   * {
     box-sizing: border-box;
+    font-family: 'Source Sans Pro', sans-serif;
```
kseniaSs, line 15: It's not a global styles? Why is this in app.vue?

neSpecc (approved):

GeekaN2 (approved):

PR #22, opened 2023-07-26
Fix build for layout

kseniaSs:
Fix npm run build:
Add .env.types
Remove layout from default.vue

### DIFF
--- a/src/vite-env.d.ts
+++ b/src/vite-env.d.ts
@@ -1,11 +1,16 @@
 // / <reference types="vite/client" />
+import 'vue-router';
+
+// To ensure it is treated as a module, add at least one `export` statement
+export {};
+
 declare module '*.vue'
 
 declare module 'vue-router' {
-    interface RouteMeta {
-      /**
-       * Layout like component wrapper
-       */
-      layout?: string
-    }
+  interface RouteMeta {
+    /**
+     * Layout like component wrapper
+     */
+    layout?: string
   }
+}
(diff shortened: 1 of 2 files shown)


src/vite-env.d.ts
```diff
@@ -1,11 +1,16 @@
 // / <reference types="vite/client" />
+import 'vue-router';
```
slaveeks, line 2: Does it have to be imported here?

slaveeks: Maybe we can omit it?

src/vite-env.d.ts
```diff
@@ -1,11 +1,14 @@
 // / <reference types="vite/client" />
+// To ensure it is treated as a module, add at least one `export` statement
+export {};
```
slaveeks, line 3: Can we omit it after removing module `import` statement?

kseniaSs: Without this {} , we get router types error, doesn't it?

```
src/application/router/index.ts:1:10 - error TS2305: Module '"vue-router"' has no exported member 'createRouter'.

1 import { createRouter, createWebHistory } from 'vue-router';
           ~~~~~~~~~~~~

src/application/router/index.ts:1:24 - error TS2305: Module '"vue-router"' has no exported member 'createWebHistory'.

1 import { createRouter, createWebHistory } from 'vue-router';
```

slaveeks: yes, you are right)

slaveeks (approved):

neSpecc (approved):

PR #24, opened 2023-07-28
Fixed building, created TabProps.ts file with TabProps interface

slaveeks:
Created `TabProps.ts` file with TabProps interface type to fix building


src/presentation/components/ui/Tab.vue
neSpecc, line 1: put them into the `/components/tabs/` folder

slaveeks: done

neSpecc (approved):

GeekaN2 (approved):

PR #41, opened 2023-07-30
feat: app state, subscribable store, user panel, refresh session

neSpecc:
## Auth

- If we have stored refresh token, auth session will be refreshed
- If user is authenticated, show user picture in a header

image

## Reactivity

- new concept "App State" — is a read-only combination of stores. It's created on Domain Level (since it creates Stores and knows about them). Application Services will have access to it. So they can react on store changes.
- SubscribableStore allows to watch data change. Any store could extend it.
- base class for Repository added

### DIFF
--- a/.architecture.eslintrc
+++ b/.architecture.eslintrc
@@ -42,8 +42,18 @@
         "mode": "file"
       },
       {
+        "type": "Abstract Repository",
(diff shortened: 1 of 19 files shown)


slaveeks (commented):
Please, fix linter errors

slaveeks (approved):

PR #44, opened 2023-08-27
Added global styles for texts

kseniaSs:
Added global styles for texts

I'm not sure about style's folder

### DIFF
--- a/src/presentation/pages/Home.vue
+++ b/src/presentation/pages/Home.vue
@@ -1,7 +1,20 @@
 <template>
-  <div>
-    <h1>Recent Notes</h1>
-    <p>This page will contain your Notes you recently worked with </p>
+  <div class="page__block">
+    <h1 class="page__title">
+      Recent Notes
+    </h1>
+    <h2 class="page__title--2">
+      What does it mean «block-styled editor»
+    </h2>
+    <h3 class="page__title--3">
+      What does it mean «block-styled editor»
+    </h3>
+    <p class="page__content">
+      Workspace in classic editors is made of a single contenteditable element, used to create different HTML markups. Editor.js workspace consists of separate Blocks: paragraphs, headings, images, lists, quotes, etc. Each of them is an independent contenteditable element (or more complex structure) provided by Plugin and united by Editor's Core.
+    </p>
+    <p class="page__content-small">
+      Workspace in classic editors is made of a single contenteditable element, used to create different HTML markups. Editor.js workspace consists of separate Blocks: paragraphs, headings, images, lists, quotes, etc. Each of them is an independent contenteditable element (or more complex structure) provided by Plugin and united by Editor's Core.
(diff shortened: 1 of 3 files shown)


src/presentation/pages/Home.vue
```diff
@@ -1,7 +1,20 @@
 <template>
-  <div>
-    <h1>Recent Notes</h1>
-    <p>This page will contain your Notes you recently worked with </p>
+  <div class="page__block">
```
neSpecc, line 2: 1. If you're using BEM — you're declared an element, but the element should be included in a Block. 
2. It's better to place styles for blocks and elements inside their component. See [Colocation of inherently coupled concerns](https://vuejs.org/#what-about-separation-of-concerns).  
3. Styles should be applied to the Note (with editor), not a static page (Home)
4. If you want to move note styles to the `typography.pcss` to allow use them somewhere else except the editor (where?), maybe it would be better to use CSS Modules instead of bem? Because for now styles are coupled with their classes, so we will need to use the same classes (`.page__title--2`) in different components?

src/presentation/styles/typography.pcss
```diff
@@ -1,35 +1,74 @@
...
+        margin: var(--spacing-mm) 0 var(--spacing-xxs);
+    }
+
+    h3 {
+        @apply --text-heading-3;
+        padding: 0;
+        margin: var(--spacing-xs) 0 var(--spacing-very-x);
+    }
+
+    .ce-paragraph {
```
neSpecc, line 65: its better to place editor styles at the Editor component

src/presentation/styles/typography.pcss
```diff
@@ -1,35 +1,74 @@
...
 
   --text-small {
+      color: var(--color-text-main);
       font-weight: 400;
       font-size: 0.875rem;
       line-height: 120%;
       letter-spacing: 0.14px;
   }
+
+    .block {
```
neSpecc, line 41: where did it come from?

src/presentation/styles/typography.pcss
```diff
@@ -1,35 +1,74 @@
...
   --text-heading-3 {
+      color: var(--color-text-heading);
       font-weight: 700;
       font-size: 1.375rem;
       line-height: 120%;
       letter-spacing: 0.1px;
   }
 
   --text-body {
+      color: var(--color-text-main);
```
neSpecc, line 27: why do we need to override color?

src/presentation/styles/typography.pcss
kseniaSs: I'm not sure, but styling global styles inside components is a good pattern?

neSpecc: I don't think it is a global style in that case. It's a style of the Editor. There won't be a `.ce-paragraph` outside the `<Editor>` component

PR #50, opened 2023-09-18
Note settings refactoring

kseniaSs:
Refactor noteSetting's services/repositories. 
Move functions from note.ts/note.service/note.repository.


src/application/services/useNoteSettings.ts
```diff
@@ -0,0 +1,47 @@
...
+   * NoteSettings ref
+   */
+  const noteSettings = ref<NotesSettings | null>(null);
+
+  /**
+   * Get note settings
+   *
+   * @param publicId - Note publicId
+   */
+  const loadSettings = async (publicId: string): Promise<void> => {
```
neSpecc, line 37: it's better not to use the "public" prefix, because from the client-perspective there is only "id" property. It doesn't know about internal id which is a backend feature.
```suggestion
  const loadSettings = async (id: string): Promise<void> => {
```

src/infrastructure/storage/noteSettings.ts
neSpecc, line 1: I think we can get rid of Note Setting Store at the moment. We can add it later if we will need it.

src/application/services/useNoteSettings.ts
```diff
@@ -0,0 +1,47 @@
...
+   * NoteSettings ref
+   */
+  const noteSettings = ref<NotesSettings | null>(null);
+
+  /**
+   * Get note settings
+   *
+   * @param id - Note id
+   */
+  const loadSettings = async (id: string): Promise<void> => {
```
neSpecc, line 37: it's better to use the `NoteId` type instead of `string`

src/domain/noteSettings.repository.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export default interface NoteSettingsRepositoryInterface {
+
+  /**
+   * Returns NotesSettings by publicId
+   *
+   * @param publicId - note publicId
+   * @returns NotesSettings | null - NotesSettings instance
+   */
+  getNotesSettingsById(publicId: string): Promise<NotesSettings | null>;
```
neSpecc, line 14: Here and following: publicId -> id

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,65 @@
...
+   */
+  public async getNotesSettingsById(publicId: string): Promise<NotesSettings | null> {
+    const notesSettingsData = await this.noteStorage.getNotesSettingsById(publicId);
+
+    /**
+     * If notesSettings data in storage exists
+     */
+    if (notesSettingsData) {
+      return notesSettingsData;
+    }
```
neSpecc, line 46: these lines have no sense, since we have removed noteSettings storage

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * If note data in API payload exists
+     */
+    if (notesSettings) {
+      /**
+       * Insert note to storage
+       */
+      await this.noteStorage.updateNotesSettings(notesSettings);
+    }
```
neSpecc, line 61: same

src/application/services/useNoteSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+   * @param id - Note id
+   */
+  const loadSettings = async (id: NoteId): Promise<void> => {
+    noteSettings.value = await noteSettingsService.getNotesSettingsById(id);
+  };
+
+
+  return {
+    noteSettings,
+    loadSettings,
```
neSpecc, line 45: lets call them `settings` and `load`

src/domain/noteSettings.repository.interface.ts
```diff
@@ -0,0 +1,15 @@
...
+ */
+export default interface NoteSettingsRepositoryInterface {
+
+  /**
+   * Returns NotesSettings by  id
+   *
+   * @param  id - note  id
+   * @returns NotesSettings | null - NotesSettings instance
+   */
+  getNotesSettingsById(id: string): Promise<NotesSettings | null>;
```
neSpecc, line 14:
```suggestion
  getNotesSettingsById(id: NoteId): Promise<NotesSettings | null>;
```

neSpecc: I think it should not return `null`. If an incorrect id will be passed, there will be 404 exception thrown

src/domain/noteSettings.service.ts
```diff
@@ -0,0 +1,33 @@
...
+   */
+  constructor(noteSettingsRepository: NoteSettingsRepository) {
+    this.noteSettingsRepository = noteSettingsRepository;
+  }
+
+  /**
+   * Get notesSettings
+   *
+   * @param publicId - Note publicId
+   * @returns { NotesSettings | null } - Note data
```
neSpecc, line 26:
```suggestion
```

src/domain/noteSettings.service.ts
```diff
@@ -0,0 +1,33 @@
...
+   * @param noteSettingsRepository - Note repository instance
+   */
+  constructor(noteSettingsRepository: NoteSettingsRepository) {
+    this.noteSettingsRepository = noteSettingsRepository;
+  }
+
+  /**
+   * Get notesSettings
+   *
+   * @param publicId - Note publicId
```
neSpecc, line 25: do not use the "public" prefix

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,66 @@
+import type NoteSettingsRepositoryInterface from '@/domain/noteSettings.repository.interface';
+import type NotesSettings from '@/domain/entities/NotesSettings';
+import type NoteStorage from '@/infrastructure/storage/note';
+import type NotesApiTransport from '@/infrastructure/transport/notes-api';
+import type { NoteId } from '@/domain/entities/Note';
+import type { GetNotesSettingsResponsePayload } from '@/infrastructure/transport/notes-api/types/GetNoteResponsePayload';
+
+/**
+ * Note repository
```
neSpecc, line 9:
```suggestion
 * Note settings repository
```

src/infrastructure/noteSettings.repository.ts
neSpecc: Because there should be a clear reason to add a cache, and for now we don't have that reason.

src/application/services/useNoteSettings.ts
```diff
@@ -0,0 +1,48 @@
...
+
+
+/**
+ * Application service for working with the Note settings
+ */
+export default function (): UseNoteSettingsComposableState {
+  /**
+   * NoteSettings ref
+   */
+  const note = ref<NotesSettings | null>(null);
```
slaveeks, line 31: Why note?

src/domain/noteSettings.repository.interface.ts
```diff
@@ -0,0 +1,17 @@
...
+export default interface NoteSettingsRepositoryInterface {
+
+  /**
+   * Returns NotesSettings by  id
+   *
+   * @param  id - note  id
+   * @throws Will throw an error if id is not found.
+   * @returns NotesSettings - NotesSettings instance
+   */
+  getNotesSettingsById(id: NoteId): Promise<NotesSettings>;
```
neSpecc, line 16:
```suggestion
  getNoteSettingsById(id: NoteId): Promise<NoteSettings>;
```

src/domain/noteSettings.repository.interface.ts
```diff
@@ -0,0 +1,17 @@
+import type NotesSettings from '@/domain/entities/NotesSettings';
```
neSpecc, line 1:
```suggestion
import type NoteSettings from '@/domain/entities/NoteSettings';
```

src/domain/noteSettings.repository.interface.ts
```diff
@@ -0,0 +1,17 @@
...
+import type { NoteId } from './entities/Note';
+
+/**
+ * Repository interface describes the methods that required by domain for its business logic implementation
+ */
+export default interface NoteSettingsRepositoryInterface {
+
+  /**
+   * Returns NotesSettings by  id
```
neSpecc, line 10:
```suggestion
   * Returns setting for a note with passed id
```

src/domain/noteSettings.service.ts
```diff
@@ -0,0 +1,39 @@
...
+   * Note Service constructor
+   *
+   * @param noteSettingsRepository - Note repository instance
+   */
+  constructor(noteSettingsRepository: NoteSettingsRepository) {
+    this.noteSettingsRepository = noteSettingsRepository;
+  }
+
+  /**
+   * Get notesSettings
```
neSpecc, line 24:
```suggestion
   * Returns setting for a note with passed id
```

src/domain/noteSettings.service.ts
```diff
@@ -0,0 +1,39 @@
...
+   *
+   * @param id - Note id
+   */
+  public async getNotesSettingsById(id: NoteId): Promise<NotesSettings | null> {
```
neSpecc, line 28:
```suggestion
  public async getNoteSettingsById(id: NoteId): Promise<NoteSettings | null> {
```

src/domain/noteSettings.service.ts
```diff
@@ -0,0 +1,39 @@
...
+    let result;
+
+    try {
+      result = await this.noteSettingsRepository.getNotesSettingsById(id);
+    } catch (error) {
+      throw new Error(`404 Not Found: We couldn't find ${id} note.`);
+    }
```
neSpecc, line 35: I think, error handling and ui message should be at the presentation layer. Also you need to check the error for `NotFoundError` instance. 

Rough example:

```ts
/**
 * Get note settings
 *
 * @param id - Note id
 */
 const load = async (id: NoteId): Promise<void> => {
   try { 
     noteSettings.value = await noteSettingsService.getNotesSettingsById(id);
   } catch (error){
     if (error instanceof NotFoundError) {
       router.replace({
         name: "404"
       });
       
       return;
     }
     
     // other errors should be thrown down
     throw error;
   }
 };
```

I think error handling could be done via a separate PR.

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+    this.transport = notesApiTransport;
+  }
+
+  /**
+   * Get NotesSettings by note ID
+   *
+   * @param id - Note id
+   * @returns { NotesSettings } - NotesSettings instance
+   */
+  public async getNotesSettingsById(id: NoteId): Promise<NotesSettings> {
```
neSpecc, line 39:
```suggestion
  public async getNoteSettingsById(id: NoteId): Promise<NoteSettings> {
```

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+    return await this.transport.get<GetNotesSettingsResponsePayload>('/note/' + id + '/settings');
```
neSpecc, line 40:
```suggestion
    return await this.transport.get<GetNoteSettingsResponsePayload>('/note/' + id + '/settings');
```

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,42 @@
+import type NoteSettingsRepositoryInterface from '@/domain/noteSettings.repository.interface';
+import type NotesSettings from '@/domain/entities/NotesSettings';
+import type NoteStorage from '@/infrastructure/storage/note';
+import type NotesApiTransport from '@/infrastructure/transport/notes-api';
+import type { NoteId } from '@/domain/entities/Note';
+import type { GetNotesSettingsResponsePayload } from '@/infrastructure/transport/notes-api/types/GetNoteResponsePayload';
```
neSpecc, line 6:
```suggestion
import type { GetNoteSettingsResponsePayload } from '@/infrastructure/transport/notes-api/types/GetNoteResponsePayload';
```

neSpecc: probably, could be moved to the separate file

neSpecc (changes requested):

src/infrastructure/noteSettings.repository.ts
```diff
@@ -0,0 +1,42 @@
...
+export default class NoteSettingsRepository implements NoteSettingsRepositoryInterface {
+  /**
+   * Transport instance
+   */
+  private readonly transport: NotesApiTransport;
+
+  /**
+   * Note storage
+   */
+  private noteStorage: NoteStorage;
```
neSpecc, line 20: seems irrelevant

src/infrastructure/storage/note.ts
```diff
@@ -41,24 +41,24 @@ export default class NoteStorage {
...
   /**
-   * Update notesSettings
+   * Update noteSettings
    *
-   * @param notesSettings - notesSettings to update
+   * @param noteSettings - noteSettings to update
    */
-  public async updateNotesSettings(notesSettings: NotesSettings): Promise<void> {
-    console.log('Update notesSettings', notesSettings);
+  public async updateNoteSettings(noteSettings: NoteSettings): Promise<void> {
```
neSpecc, line 60: unused code, better remove it

slaveeks (approved):

neSpecc (approved):

PR #51, opened 2023-10-16
chore(build): eslint fix

neSpecc:
This PR fixes all ESLint errors

##### src/application/services/useLayout.ts

```bash
  20:20  error  Unexpected any value in conditional. An explicit comparison or type cast is required  @typescript-eslint/strict-boolean-expressions
```

##### src/domain/note.service.ts

```bash
  27:1  warning  Missing JSDoc @returns type  jsdoc/require-returns-type
```

##### src/infrastructure/note.repository.ts

```bash
   37:1  warning  Missing JSDoc @returns type                                                           jsdoc/require-returns-type
   80:9  error    Unexpected any value in conditional. An explicit comparison or type cast is required  @typescript-eslint/strict-boolean-expressions
  114:9  error    Unexpected object value in conditional. The condition is always true                  @typescript-eslint/strict-boolean-expressions
```

##### src/infrastructure/transport/notes-api/index.ts

```bash
  24:1  warning  @param "endpoint" does not match an existing function parameter  jsdoc/check-param-names
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,7 +1,7 @@
 {
   "name": "notes.web",
   "version": "0.0.1",
-  "lockfileVersion": 2,
+  "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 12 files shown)


slaveeks:
image

slaveeks (approved):

kseniaSs (approved):

PR #68, opened 2023-10-23
Added Button to Logout in Settings

shvkhzod:
- Changed useOAuth.ts to useAuth.ts
- Created AuthLogoutEvent to notify other listeners about the logout
- Added removeUser method to user.repository to clear the user data once user logs out
- Added word logout to en.json


slaveeks:
Fix code style, you have a lot of unnecessary blank lines

src/presentation/components/button/LogOutBtn.vue
neSpecc, line 1: I think, it's better to use an existing `Button` component on a page instead of creating the new one. Let's treat our `components` folder like a UI-Kit.

src/presentation/components/button/LogOutBtn.vue
```diff
@@ -0,0 +1,29 @@
+<template>
+<Button 
+  text="Log out"
```
neSpecc, line 3: pls display text using i18n

src/presentation/components/button/LogOutBtn.vue
```diff
@@ -0,0 +1,29 @@
...
+  :icon="IconUnlink"
+  type="primary"
+  @click="logOut"
+  />
+</template>
+
+<script setup lang="ts">
+import { IconUnlink } from '@codexteam/icons';
+import Button from './Button.vue';
+import { authService } from '@/domain/index';
```
neSpecc, line 13: Accessing a domain level from Presentation level is prohibited. ESLint should warn about it, try `yarn lint`.

You can use Application Service for that reason.

I'd suggest renaming `useOAuth()` to `useAuth()` and put a logout method in there.

src/application/services/useAuth.ts
```diff
@@ -59,3 +66,19 @@ export default function useOAuth(): UseOAuthComposableState {
...
 }
+
+/**
+ * Methods for working with Auth tokens
+ */
+export function useAuth(): UseAuthComposableState {
+  /**
+   * Deletes the refresh token in the local storage
+   */
+  async function logoutTheUser(): Promise<void> {
```
neSpecc, line 77: I guess we can just add this method to existing composable, no need to create another one. Just rename useOAuth to useAuth

src/application/services/useAuth.ts
```diff
@@ -54,8 +58,16 @@ export default function useOAuth(): UseOAuthComposableState {
       }
     });
   }
+  /**
+   *Logs out the user by deleting the refresh token in local strorage
+   */
+  async function logoutTheUser(): Promise<void> {
```
neSpecc, line 64:
```suggestion
  async function logout(): Promise<void> {
```

src/presentation/pages/Settings.vue
```diff
@@ -1,14 +1,36 @@
...
+const router = useRouter();
+const { logoutTheUser } = useAuth();
 
+/**
+ * Function to logout the user by deleting the refresh  token from the local storage and redirecting to the main page
+ */
+async function logout() {
+  await logoutTheUser().then(() => {
+    router.push({ path: '/' }).then(() => {
+      window.location.reload();
```
neSpecc, line 30: why do we need to hard reload a page? It's better to "unauthorize" auth-related staff. At least `notesApiTransport`. Just like we do after authorisation but with opposite actions. Check the `todo` in auth.sevice.ts / logout method

shvkhzod: Even after I removed the token from the local storage, the change was not apparent until the hard reload. I could access profile settings, profile avatar was still there, etc

neSpecc (commented):
You also need to unauthorise NotesApiTransport instance, check `infrastructure/index.ts`.

There is also one suggestion left. Check Files Changed page.

src/infrastructure/index.ts
```diff
@@ -74,6 +75,15 @@ export function init(noteApiUrl: string, eventBus: EventBus): Repositories {
...
   });
+  /**
+   * When we got unauthorized
+   */
+  eventBus.addEventListener(AUTH_LOGOUT_EVENT_NAME, () => {
+    /**
+     * Tell API transport to continue working in anonymous mode (send waiting requests without auth)
+     */
+    notesApiTransport.continueAnonymous();
+  });
```
neSpecc, line 86: probably, you also need removing the refresh token from authStore

neSpecc:
Please, update a branch

neSpecc (approved):

MedVAL-na (approved):

PR #75, opened 2023-11-06
feature: adding list of tools in /marketplace

MedVAL-na:
Task: "Add list of tools to /marketplace"

Done:

-  created /marketplace page and route to it  (Marketplace.vue, routes.ts)
-  added editor tools to entities(EditorTool.ts)
-  created interface for repository of tools(tools.repository.interface.ts)
-  created  repository of tools and implemented method of getting list of tools from backend(tools.repository.ts)
-  created editor tool service class(tools.service.ts)
-  added tool repository to initialization of repositories function (index.ts)
-  added list of tools in useAppState(useAppState.ts)

I have no idea why list of tools doesn't appear at page /marketplace.

### DIFF
--- a/src/application/router/routes.ts
+++ b/src/application/router/routes.ts
@@ -4,6 +4,7 @@ import Landing from '@/presentation/pages/Landing.vue';
 import Settings from '@/presentation/pages/Settings.vue';
 import NoteSettings from '@/presentation/pages/NoteSettings.vue';
 import ErrorPage from '@/presentation/pages/Error.vue';
+import Marketplace from '@/presentation/pages/Marketplace.vue';
 import type { RouteRecordRaw } from 'vue-router';
 
 // Default production hostname for homepage. If different, then custom hostname used
@@ -54,6 +55,10 @@ const routes: RouteRecordRaw[] = [
     props: route => ({
       id: String(route.params.id),
     }),
+  },
+  {
+    path: `/marketplace/`,
+    component: Marketplace,
   },
   /**
    * 404 page
--- a/src/domain/entities/EditorTool.ts
+++ b/src/domain/entities/EditorTool.ts
@@ -34,9 +34,10 @@ export interface EditorTool {
      * Source of the tool to get it's code
      */
     source: {
-        /**
-         * Tool URL in content delivery network
-         */
+    /**
+     * Tool URL in content delivery network
+     */
+
         cdn?: string;
     }
 }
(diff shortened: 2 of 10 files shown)


src/domain/entities/EditorTool.ts
```diff
@@ -0,0 +1,42 @@
...
+
+    /**
+     * Source of the tool to get it's code
+     */
+    source: {
+        /**
+         * Tool URL in content delivery network
+         */
+        cdn?: string;
+    }
```
neSpecc, line 41: bad code style

neSpecc: install Editor Config plugin so it will respect `.editorconfig` file

src/infrastructure/tools.repository.ts
```diff
@@ -0,0 +1,44 @@
...
+   */
+  constructor(store: EditorToolStore, notesApiTransport: NotesApiTransport) {
+    super(store);
+
+    this.transport = notesApiTransport;
+  }
+
+
+  /**
+   *
```
neSpecc, line 30: ?

src/application/router/routes.ts
```diff
@@ -55,6 +56,10 @@ const routes: RouteRecordRaw[] = [
       id: String(route.params.id),
     }),
   },
+  {
+    path: `/marketplace/`,
```
neSpecc, line 60:
```suggestion
    path: `/marketplace`,
```

src/infrastructure/tools.repository.ts
```diff
@@ -0,0 +1,43 @@
...
+
+    this.store.setEditorTools(response);
+  };
+
+  /**
+   * Load editor tools data and put it to the storage
+   */
+  public getEditorTool(): EditorTool | null  {
+    return this.store.getEditorTool();
+  }
```
neSpecc, line 42: there is some inconsistence in naming:
- tools
- EditorTool
- EditorTools

pick up the right one and use it everywhere

src/infrastructure/storage/tools.ts
```diff
@@ -0,0 +1,31 @@
...
+  /**
+   * User data
+   */
+  editorTool: EditorTool | null;
+};
+
+/**
+ * Store for the editor tools
+ */
+export class EditorToolStore extends SubscribableStore<EditorToolStoreData> {
```
neSpecc, line 17: I think, you can get rid of the store for this task. Just load tools from transport every time.

Store is used when we need to keep some data in memory, but there is no reason for that at the moment. We can add it later if needed.

Subscribable Store — is a kind of store that empowers reactivity to the stored data.

PR #78, opened 2023-11-22
Added getting note access rights from note.api

kloV148:
This PR is related to pr122 in notes.api
Changes:
- added NoteAccessRights entity
- GetNoteResponsePayload now have { note: Note, accessRights: NoteAccessRights } structure
- NoteRepositoryInterface, NoteRepository, NoteService, UseNote updated to work with new note payload
- Removed cases when trying to get a note by its hostname could return null (because when the note is not found,   a    NotFoundError should be thrown)

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -96,7 +96,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     /**
      * @todo try-catch domain errors
      */
-    note.value = await noteService.getNoteById(id);
+    note.value = (await noteService.getNoteById(id)).note;
   };
 
   /**
@@ -136,7 +136,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    * Get note by custom hostname
    */
   const resolveHostname = async (): Promise<void> => {
-    note.value = await noteService.getNoteByHostname(location.hostname);
+    note.value = (await noteService.getNoteByHostname(location.hostname)).note;
   };
 
   onMounted(() => {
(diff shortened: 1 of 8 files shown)


src/application/services/useNote.ts
```diff
@@ -96,7 +96,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     /**
      * @todo try-catch domain errors
      */
-    note.value = await noteService.getNoteById(id);
+    note.value = (await noteService.getNoteById(id)).note;
```
neSpecc, line 99: at this moment, access right are ignored. Let's set them to a separate variable next to `note` so it could be available in a component. Then, use this variable to pass `readOnly: true` to the Editor.js if `canEdit=false`

kloV148:
Now if the user is not the creator of the note, the editor switches to read only mode. 
UseNote send NoteAccessRights instance. In Note the canEdit flag is passed to the editor using props. If canEdit=False, then the editor will be in read only mode

src/application/services/useNote.ts
```diff
@@ -87,6 +104,13 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    */
   const router = useRouter();
 
+  /**
+   * Access rights for currently opened note
+   *
+   * Create new NoteAccessRights instance when new note is created
+   */
+  const accessRights = ref<NoteAccessRights | null>(currentId.value == null ? createAccessRights() : null);
```
neSpecc, line 112: I think you can simply declare `canEdit` ref with "true" by default. So you won't need `createAccessRights()` method with looks redundant

neSpecc (approved):

GoldenJaden (approved):

PR #82, opened 2024-01-06
Typescript intellisence for vite environment variables in vue file

iamgabrielsoft:
Typescript exposes vite environmental variables to the code for intellisence easy pickup

@neSpecc 

Screenshot (93)

### DIFF
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -1,0 +1,7 @@
+
+/**
+ * Intellisense for Typescript
+ */
+interface ImportMetaEnv {
+  VITE_API_URL: string;
+}
--- a/src/env.d.ts
+++ b/src/env.d.ts
@@ -1,7 +1,26 @@
 
 /**
- * Intellisense for Typescript
+/**
+ * @description - Typescript Intellisense for VITE Environmental variables
+ * Add VITE local variables here
+ * Add VITE local variables here
  */
 interface ImportMetaEnv {
+  /**
+   * Notex hosted backend server url from environmental variable file(.env)
+   */
   VITE_API_URL: string;
+  /**
+   * Notex google oauth url from environmental variable file (.env )
+   */
+  VITE_GOOGLE_OAUTH_URL: string;
+  /**
+   * local backend server url from environmental variable file (.env)
+   */
+  VITE_API_LOCAL_URL: string;
+  /**
+   * local google oauth url from environmental variable file (.env)
+   * Add your preferred google oauth url
+   */
+  VITE_GOOGLE_LOCAL_OAUTH_URL: string;
 }
--- a/src/application/services/useAuth.ts
+++ b/src/application/services/useAuth.ts
@@ -19,7 +19,7 @@ export default function useAuth(): UseOAuthComposableState {
   /**
    * Google OAuth URL
    */
-  const loginUrl = import.meta.env.VITE_GOOGLE_OAUTH_URL as string;
+  const loginUrl = import.meta.env.VITE_GOOGLE_OAUTH_URL;
 
   /**
    * Id of the postMessage callback
--- a/src/application/services/useAuth.ts
+++ b/src/application/services/useAuth.ts
@@ -59,7 +59,7 @@ export default function useAuth(): UseOAuthComposableState {
     });
   }
   /**
-   *Logs out the user by deleting the refresh token in local strorage
+   *Logs out the user by deleting the refresh token in local storage
    */
   async function logout(): Promise<void> {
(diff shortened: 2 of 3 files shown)


TatianaFomina:
Please add description to the pull request

src/env.d.ts
```diff
@@ -0,0 +1,7 @@
+
+/**
+ * Intellisense for Typescript
+ */
+interface ImportMetaEnv {
+  VITE_API_URL: string;
```
neSpecc, line 6: Let's add a JSDocs here

iamgabrielsoft: done

TatianaFomina (approved):

src/env.d.ts
neSpecc: I can't see updates

iamgabrielsoft: I've updated it

TatianaFomina: I suppose, @neSpecc suggests adding comment describing `VITE_API_URL` purpose

iamgabrielsoft: noted @TatianaFomina

iamgabrielsoft: @TatianaFomina I've updated, I don't know if @neSpecc meant otherwise

src/env.d.ts
```diff
@@ -0,0 +1,26 @@
+
+/**
+/**
+ * @description - Typescript Intellisense for VITE Environmental variables
+ * Add VITE local variables here
+ * Add VITE local variables here
```
neSpecc, line 6:
```suggestion
```

src/env.d.ts
```diff
@@ -0,0 +1,26 @@
+
+/**
```
neSpecc, line 2:
```suggestion
```

src/env.d.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * @description - Typescript Intellisense for VITE Environmental variables
+ * Add VITE local variables here
+ * Add VITE local variables here
+ */
+interface ImportMetaEnv {
+  /**
+   * Notex hosted backend server url from environmental variable file(.env)
```
neSpecc, line 10:
```suggestion
   * NoteX API server URL
```

src/env.d.ts
```diff
@@ -0,0 +1,26 @@
...
+   */
+  VITE_API_URL: string;
+  /**
+   * Notex google oauth url from environmental variable file (.env )
+   */
```
neSpecc, line 15:
```suggestion

  /**
   * Google OAuth redirect URL
   */
```

src/env.d.ts
```diff
@@ -0,0 +1,26 @@
...
+  VITE_GOOGLE_OAUTH_URL: string;
+  /**
+   * local backend server url from environmental variable file (.env)
+   */
+  VITE_API_LOCAL_URL: string;
```
neSpecc, line 20: we don't have such env variable

src/env.d.ts
```diff
@@ -0,0 +1,26 @@
...
+  /**
+   * local google oauth url from environmental variable file (.env)
+   * Add your preferred google oauth url
+   */
+  VITE_GOOGLE_LOCAL_OAUTH_URL: string;
```
neSpecc, line 25: same

neSpecc (changes requested):

src/env.d.ts
iamgabrielsoft: this has been updated on the latest push

src/env.d.ts
iamgabrielsoft: its has being added

neSpecc: why? PR description says nothing about it. The reason of this vars is unclear, since we already have them. They are not used in a code.

neSpecc:
Outdated

neSpecc (changes requested):
No updates

src/env.d.ts
iamgabrielsoft: I will comment those out, probably for my use alone

iamgabrielsoft (commented):
certain env variables were added due to the backend server being down and testing of Google OAuth locally(possibilities).

in any case, those will be commented out

neSpecc:
> certain env variables were added due to the backend server being down and testing of Google OAuth locally(possibilities).
> 
> in any case, those will be commented out

you can override existing env variables in our .env file

iamgabrielsoft:
please let's move here PR #123

PR #86, opened 2024-01-10
Seperate Base and Accent Colors

shvkhzod:
A feature to handle themes, specifically setting base and accent colors of 4 themes(Default, Crimson,Violet and Red) seperately.
For example, page background uses base colors while many components use accent Colors, so a user should be able to set different styles, i,e: Default Base Colors with Red Accent Colors.


src/application/services/themeService.ts
```diff
@@ -0,0 +1,41 @@
+// ThemeService.ts
```
TatianaFomina, line 1: commented out line

src/application/services/themeService.ts
```diff
@@ -0,0 +1,41 @@
...
+ * Changes Accent theme
+ *
+ * @param accentThemeName - the name of Accent theme
+ */
+const setAccentTheme = (accentThemeName:string):void => {
+  currentAccentTheme.value = accentThemeName;
+  localStorage.setItem('accentTheme', accentThemeName);
+
+
+  // applyAccentTheme();
```
TatianaFomina, line 16: commented out line

src/presentation/components/webComponents/button.ts
TatianaFomina, line 1: is this file needed?

TatianaFomina (changes requested):
1. There are some color values duplicates in `themes.pcss`. Probably you should create separate file `colors.pcss` or `variables.pcss` where all the available color variables will be present. They will not have `base` or `accent` prefix, name them like `--color-violet-bgPrimary`

2. You should not update css variables values with javascript. Once theme value changes, just add relevant classname to the root element of the page. 
`themes.pcss` should selectors for all available theme classes and overwrite variables values for them using abovementioned `colors.pcss`

src/presentation/styles/theme-classic.pcss
```diff
@@ -0,0 +1,26 @@
+.theme-base-pure {
```
TatianaFomina, line 1: why not `.theme-base-classic`?

src/presentation/styles/colors.pcss
```diff
@@ -32,6 +32,98 @@
    */
   --color-line: var(--color-gray-superlight);
   --color-delimiter-light: var(--color-transparent);
+
+  /**
+   * Classic Dark Colors
+   */
+   --bg-primary-classic: #242527;
```
TatianaFomina, line 39: Maybe add `color` prefix/postfix to color variables to make their purpose more clear?

src/application/services/themeService.ts
```diff
@@ -0,0 +1,37 @@
+import { ref } from 'vue';
+
+const currentBaseTheme = ref((localStorage.getItem('baseTheme') ?? '') || 'theme-base-pure');
+const currentAccentTheme = ref((localStorage.getItem('accentTheme') ?? '') || 'theme-accent-classic');
```
TatianaFomina, line 4: according to @neSpecc's architecture, you can't just access local storage from application service 🥲

you need to create a storage, then access it from repository, access repository from domain service, and access service from application service. 

you can examine how storing user data is implemented to understand better

src/application/services/themeService.ts
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Changes Accent theme
+ *
+ * @param accentThemeName - the name of Accent theme
+ */
+const setAccentTheme = (accentThemeName:string):void => {
+  currentAccentTheme.value = accentThemeName;
```
TatianaFomina, line 11: I think you can add relevant css class to body element here in setter function instead of watching changes in App.vue

src/presentation/components/webComponents/button.ts
TatianaFomina, line 1: is this file needed?

src/presentation/components/tabs/Tab.vue
```diff
@@ -20,11 +21,22 @@
...
  * Tab component props
  */
 defineProps<Tab>();
+const theme = computed(() => themeService.currentBaseTheme.value);
+
+onMounted(() => {
+  var all = document.getElementsByClassName('tab');
+
+  for (var i = 0; i < all.length; i++) {
+    (all[i] as HTMLElement).style.color = `var(--base-${theme.value.toLowerCase()}-text)`;
```
TatianaFomina, line 37: why not just set it via css?
```
color: var(--text-base);
```

src/presentation/styles/colors.pcss
neSpecc: I don't think so, its better to stay with the same naming as Figma

src/presentation/styles/theme-classic.pcss
shvkhzod: it was a name from Figma

neSpecc:
Resolved by #96

PR #104, opened 2024-02-23
Add attribute "private" to package.json

GoldenJaden:
according to yarn docs, "private:true" is needed in order to work with workspaces


kloV148 (approved):
Let's try

neSpecc (approved):

PR #110, opened 2024-02-24
Fix multiple tool loading

TatianaFomina:
Store already loaded tools in a map and do not load them the next time you open a note

### DIFF
--- a/src/application/services/useEditor.ts
+++ b/src/application/services/useEditor.ts
@@ -2,11 +2,12 @@ import type { BlockTool } from '@editorjs/editorjs';
 import Editor, { type OutputData, type API } from '@editorjs/editorjs';
 // @ts-expect-error editor plugins have no types
 import Header from '@editorjs/header';
-import type { Ref } from 'vue';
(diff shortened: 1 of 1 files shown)


src/application/services/useEditor.ts
```diff
@@ -43,51 +44,57 @@ interface UseEditorParams {
...
+export const useEditor = createSharedComposable(({ id, content, isReadOnly, onChange }: UseEditorParams): void => {
   /**
    * Editor instance
    */
   let editor: Editor | undefined;
 
+  /**
+   * Downloaded tools data
+   */
+  const downloadedTools = shallowRef<DownloadedTools>({})
```
neSpecc, line 56: should it be reactive? maybe just `Set()`?

TatianaFomina: if it's not reactive, I'm not sure, that `createSharedComposable` will work

neSpecc: you can try

PR #118, opened 2024-02-25
Fix create child note

TatianaFomina:
- Update editor once page route changes to '/note/:noteId/new'
- Replace `alert` with `console.error` in global errors handler
- Move more logic to `useEditor`

### DIFF
--- a/src/application/services/useEditor.ts
+++ b/src/application/services/useEditor.ts
@@ -156,8 +156,9 @@ export function useEditor({ id, content, isReadOnly, onChange }: UseEditorParams
   }
 
   /**
+   * Reinitializes editor instance with new data
    *
-   * @param data
+   * @param data - new data to be displayed in editor
    */
   async function refresh(data?: OutputData): Promise<void> {
     editor?.destroy();
--- a/src/application/services/useEditor.ts
+++ b/src/application/services/useEditor.ts
@@ -45,6 +45,7 @@ interface UseEditorParams {
  */
 export function useEditor({ id, content, isReadOnly, onChange }: UseEditorParams): {
   isEmpty: Ref<boolean>;
+  refresh: (data: OutputData) => void;
 } {
   /**
    * Editor instance
@@ -130,14 +131,16 @@ export function useEditor({ id, content, isReadOnly, onChange }: UseEditorParams
 
   /**
    * Initializes editorjs instance
+   *
+   * @param data
    */
-  async function mountEditor(): Promise<void> {
+  async function mountEditor(data?: OutputData): Promise<void> {
     try {
       const tools = await downloadTools(userEditorTools);
 
       editor = new Editor({
         holder: id,
-        data: content,
+        data: data,
         tools: {
           header: Header,
           ...tools,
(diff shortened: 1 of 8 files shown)


src/application/services/useEditor.ts
```diff
@@ -99,8 +155,17 @@ export function useEditor({ id, content, isReadOnly, onChange }: UseEditorParams
     }
   }
 
+  /**
+   *
+   * @param data
```
neSpecc, line 160: docs missed

src/application/services/useNote.ts
```diff
@@ -139,7 +139,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    * @param content - Note content (Editor.js data)
    * @param parentId - Id of the parent note. If null, then it's a root note
```
neSpecc, line 140: describe when it can be "undefined"

src/presentation/pages/Note.vue
```diff
@@ -40,7 +41,7 @@ const props = defineProps<{
   /**
    * Parent note id, null for root note
    */
-  parentId: string | null;
+  parentId?: string | null;
```
neSpecc, line 44: describe null/undefined difference meaning

src/presentation/pages/Note.vue
```diff
@@ -77,22 +78,28 @@ function createChildNote(): void {
...
+    /** If new child note is created, refresh editor with empty data */
+    if (props.id === null) {
+      editor.value?.refresh();
+
       useHead({
-        title: noteTitle.value,
+        title: t('note.new'),
+      });
+    } else {
+      watchEffect(() => {
```
neSpecc, line 92: watch inside watch?

src/application/services/useNote.ts
TatianaFomina: done

src/application/services/useEditor.ts
TatianaFomina: fixed

src/presentation/pages/Note.vue
TatianaFomina: null is not needed, removed

src/presentation/pages/Note.vue
TatianaFomina: removed

src/application/services/useNote.ts
```diff
@@ -49,7 +49,7 @@ interface UseNoteComposableState {
   /**
    * Creates/updates the note
    */
-  save: (content: NoteContent, parentId: NoteId | null) => Promise<void>;
+  save: (content: NoteContent, parentId: NoteId | null | undefined) => Promise<void>;
```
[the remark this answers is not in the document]
TatianaFomina: null was incorrect, only `undefined` stays for cases when note does not have parent note

src/domain/note.service.ts
```diff
@@ -47,7 +47,7 @@ export default class NoteService {
    * @param content - Note content (Editor.js data)
    * @param parentId - Id of the parent note. If null, then it's a root note
```
neSpecc, line 48:
```suggestion
   * @param parentId - Id of the parent note. If omitted, then it's a root note
```

neSpecc (approved):

PR #119, opened 2024-02-25
feature(noteSettings): revoke hash

elizachi:
When the button is pressed, a new hash is generated

image

### DIFF
--- a/src/infrastructure/transport/fetch.transport.ts
+++ b/src/infrastructure/transport/fetch.transport.ts
@@ -103,7 +103,10 @@ export default class FetchTransport {
   public async patch(endpoint: string, payload?: JSONValue): Promise<JSONValue> {
     this.headers.set('Content-Type', 'application/json');
 
-    /** If body is undefined, we send an empty body */
+    /**
+     * If the body undefined, request fails fastify validation
+     * and sends a Bad Request, so we make it empty
+     */
     const data = payload !== undefined ? payload : {};
 
     const response = await fetch(this.baseUrl + endpoint, {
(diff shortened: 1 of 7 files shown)


src/presentation/pages/NoteSettings.vue
```diff
@@ -17,6 +17,11 @@
       :title="t('noteSettings.invitationHash')"
       :placeholder="t('noteSettings.invitationHashPlaceholder')"
     />
+    <Button
+      text="Regenerate hash"
```
neSpecc, line 21: Lets use `i18n` and name `Revoke`

src/presentation/pages/NoteSettings.vue
```diff
@@ -66,6 +71,13 @@ function onClick() {
   });
 }
 
+/**
+ * Regenerate invitation hash
+ */
+async function regenerateHash() {
+  revokeHash(props.id);
```
TatianaFomina, line 78: I think, `revokeHash` should be called `regenerateHash` as well

e11sy (approved):

TatianaFomina (approved):

src/infrastructure/transport/fetch.transport.ts
```diff
@@ -103,10 +103,13 @@ export default class FetchTransport {
   public async patch(endpoint: string, payload?: JSONValue): Promise<JSONValue> {
     this.headers.set('Content-Type', 'application/json');
 
+    /** If body is undefined, we send an empty body */
```
neSpecc, line 106: Describe a problem with Fastify you're reached

neSpecc (approved):

PR #122, opened 2024-02-25
feat(ui): form field

neSpecc:
- FormField component added
- It support 3 sizes as well as Input and Button: `small`, `medium`, `large`
- Input component now has the `value` prop instead of `text`
- `v-model` added to the Input and Form Field components allowing to bind values

image

### DIFF
--- a/codex-ui/src/vue/input/Input.vue
+++ b/codex-ui/src/vue/input/Input.vue
@@ -1,22 +1,20 @@
 <template>
-  <div>
-    <input
-      :class="[$style.input, `${$style.input}--${size}`, 'text-ui-base']"
-      :value="props.text"
-      :disabled="props.disabled"
-    />
-  </div>
+  <input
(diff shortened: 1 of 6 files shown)


slaveeks (approved):

codex-ui/src/vue/input/Input.vue
```diff
@@ -29,10 +27,17 @@ const props = withDefaults(
     disabled?: boolean;
   }>(),
   {
+    value: '',
```
elizachi, line 30: Why value is empty? Why put it in here at all if it's optional?

neSpecc: It don't make a big difference, since this prop is optional. 
In TypeScript's `lib.dom.d.ts` HTMLInputElement has a `value` prop which is `string`. So it can't be `undefined`

elizachi (approved):

PR #124, opened 2024-03-02
Feat: unlink parent

kloV148:
What's done:
- now we can get parentNote form API
- button to unlink parent
- update button, that create child note

Снимок экрана 2024-03-02 в 22 38 53

### DIFF
--- a/src/application/i18n/messages/en.json
+++ b/src/application/i18n/messages/en.json
@@ -27,7 +27,9 @@
     "revokeHash": "Revoke"
   },
   "note": {
-    "new": "New Note"
+    "new": "New Note",
+    "unlink": "Unlink",
+    "createChild": "Create Child"
   },
   "noteList": {
     "buttons": {
--- a/src/application/i18n/messages/en.json
+++ b/src/application/i18n/messages/en.json
@@ -28,8 +28,8 @@
   },
   "note": {
     "new": "New Note",
-    "unlink": "Unlink",
-    "createChild": "Create Child"
+    "unlink": "Unlink from parent",
+    "createChild": "Create child"
   },
   "noteList": {
     "buttons": {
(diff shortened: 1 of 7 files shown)


src/domain/note.repository.interface.ts
```diff
@@ -7,13 +7,14 @@ import type NoteAccessRights from '@/domain/entities/NoteAccessRights';
  */
 export default interface NoteRepositoryInterface {
   /**
-   * Returns a Note and NoteAccessRights by id
+   * Returns a Note, NoteAccessRights and parent note if exists by id
```
GoldenJaden, line 10: * Returns a Note, NoteAccessRights and parent note (if exists) by id

e11sy (approved):

GoldenJaden (approved):

src/application/i18n/messages/en.json
```diff
@@ -27,7 +27,9 @@
     "revokeHash": "Revoke"
   },
   "note": {
-    "new": "New Note"
+    "new": "New Note",
+    "unlink": "Unlink from parent",
+    "createChild": "Create child"
```
neSpecc, line 32:
```suggestion
    "createChildNote": "Create child note"
```

src/presentation/pages/Note.vue
```diff
@@ -77,6 +85,19 @@ function createChildNote(): void {
...
 
+/**
+ * Unlink note from parent
+ */
+function unlinkButton(): void {
+  if (props.id === null) {
+    throw new Error('Note is Empty');
+  }
+
+  parentNote.value = undefined;
```
neSpecc, line 96: its better to move it to the useNote service and place after the domain service call to be sure that note is really unlinked

neSpecc (approved):

PR #126, opened 2024-03-04
Light Mode Switch

shvkhzod:
Added light mode colors of themes, using mode attribute applied to body element


codex-ui/dev/Playground.vue
```diff
@@ -1,7 +1,83 @@
 <template>
   <div :class="$style.playground">
     <Heading :level="1">Playground</Heading>
+    <Heading :level="3">Mode:</Heading>
+    <div :class="$style.buttons">
+      <Button
```
TatianaFomina, line 6: Maybe use radio buttons instead of regular buttons?

codex-ui/dev/Playground.vue
```diff
@@ -1,7 +1,83 @@
...
+        size="small"
+        @click="() => setMode('light')"
```
TatianaFomina, line 8: in vue it will also work like this

```suggestion
        @click="setMode('light')"
```

TatianaFomina (commented):
accent color switch does not work

codex-ui/dev/Playground.vue
neSpecc: We don't have radio components right now. I think, it's ok to use button For the playground at the moment

codex-ui/dev/Playground.vue
```diff
@@ -94,6 +170,18 @@ const buttons = [
   ['large', 'destructive'],
   ['large', 'disabled'],
 ];
+
+function setBaseColor(color: string) {
```
neSpecc, line 174:
```suggestion
function setBaseTheme(theme: string) {
```

codex-ui/dev/Playground.vue
```diff
@@ -94,6 +170,18 @@ const buttons = [
...
+  document.body.setAttribute('theme-base', color);
+}
+
+function setAccentColor(color: string) {
```
neSpecc, line 178:
```suggestion
function setAccentTheme(theme: string) {
```

codex-ui/dev/index.html
```diff
@@ -2,6 +2,7 @@
   <title>Codex UI</title>
 </head>
 <body
+  mode="dark"
```
neSpecc, line 5: I'm afraid that `mode` could be used by user's application for some other stuff, let's make our property more specific, like `color-scheme`

codex-ui/src/styles/themes/classic.pcss
```diff
@@ -1,4 +1,4 @@
-[theme-base='classic'],
+[mode='dark'][theme-base='classic'],
 [theme-accent='classic'] {
```
neSpecc, line 2: why dark mode is not specified near the `accent` selector?

codex-ui/src/styles/themes/classic.pcss
```diff
@@ -15,6 +15,23 @@
   --classic--text-solid-foreground: #f5f5f5;
 }
 
+[mode='light'],[theme-base='classic'],
+[theme-accent='classic'] {
```
neSpecc, line 19: same

codex-ui/src/styles/themes/crimson.pcss
```diff
@@ -16,6 +16,24 @@
   --crimson--text-solid-foreground: #f9f9f9;
 }
 
+[mode='light'],[theme-base='crimson'],
```
neSpecc, line 19: are you sure that there should be a `,` ?

neSpecc (changes requested):

codex-ui/dev/Playground.vue
TatianaFomina: radiobutton-like behavior is implemented incorrectly at the moment. I suggest using native radiobuttons for simplicity. Otherwise there should be a behaviour fix

neSpecc (changes requested):
Example in a playground is not working

neSpecc (approved):

TatianaFomina:
just use radio buttons.....

image

TatianaFomina:
crimson and violet accent themes not working

TatianaFomina:
Initial value not selected 

image

TatianaFomina (approved):

PR #130, opened 2024-03-16
feat(note-settings): partial note settings save

### LINKED ISSUE #139: Save note settings partially
Add ability to save settings partially on note settigns page

elizachi:
Now `note settings` can be saved without `save` button

Resolves #139


src/application/services/useNoteSettings.ts
```diff
@@ -71,7 +71,11 @@ export default function (): UseNoteSettingsComposableState {
    * @param data - Note settings data with new values
    */
   const update = async (id: NoteId, data: Partial<NoteSettings>): Promise<void> => {
-    noteSettings.value = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
+    const response = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
+
+    for (let key in noteSettings.value) {
+      key = response[key as keyof typeof noteSettings.value] as string;
+    }
```
neSpecc, line 78: seems unsafe. Iterate over response keys and check if note settings have the same key

src/presentation/pages/NoteSettings.vue
```diff
@@ -21,24 +23,14 @@
...
-      />
-    </div>
   </div>
   <div v-else>Loading...</div>
 </template>
 
 <script lang="ts" setup>
 import type { NoteId } from '@/domain/entities/Note';
-import TextEdit from '@/presentation/components/form/TextEdit.vue';
+// import TextEdit from '@/presentation/components/form/TextEdit.vue';
```
TatianaFomina, line 32: line

src/application/services/useNoteSettings.ts
```diff
@@ -22,12 +22,12 @@ interface UseNoteSettingsComposableState {
   load: (id: NoteId) => Promise<void>;
 
   /**
-   * Update note settings
+   * Update field isPublic in note settings
    *
    * @param id - note id
    * @param data - note settings data with new values
```
TatianaFomina, line 28: replace with newIsPublicValue

TatianaFomina (approved):

src/application/services/useNoteSettings.ts
```diff
@@ -65,13 +65,17 @@ export default function (): UseNoteSettingsComposableState {
...
+   * Update field isPublic in note settings
    *
    * @param id - Note id
-   * @param data - Note settings data with new values
+   * @param newIsPublicValue - new isPublic
    */
-  const update = async (id: NoteId, data: Partial<NoteSettings>): Promise<void> => {
-    noteSettings.value = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
+  const updateIsPublic = async (id: NoteId, newIsPublicValue?: boolean): Promise<void> => {
+    if (noteSettings.value) {
```
neSpecc, line 74: why do we need this statement?

src/application/services/useNoteSettings.ts
```diff
@@ -65,13 +65,17 @@ export default function (): UseNoteSettingsComposableState {
...
-   * Update note settings
+   * Update field isPublic in note settings
    *
    * @param id - Note id
-   * @param data - Note settings data with new values
+   * @param newIsPublicValue - new isPublic
    */
-  const update = async (id: NoteId, data: Partial<NoteSettings>): Promise<void> => {
-    noteSettings.value = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
+  const updateIsPublic = async (id: NoteId, newIsPublicValue?: boolean): Promise<void> => {
```
neSpecc, line 73: newIsPublicValue should not be optional

src/application/services/useNoteSettings.ts
elizachi: Because we can't use the construct like noteSettings.value?.isPublic = response.isPublic - expression may not be an optional property access

neSpecc: your current code allows a case when you call `upadteIsPublic()` but nothing happened. It seems like a mistake, calling of `upadteIsPublic()` should always lead service call.

src/presentation/pages/NoteSettings.vue
```diff
@@ -64,23 +56,19 @@ const invitationLink = computed(
...
 }
 
 /**
- * Regenerate invitation hash
+ * Change isPublic property
  */
-async function regenerateHash() {
-  revokeHash(props.id);
+async function changeAccess() {
+  if (noteSettings.value) {
```
neSpecc, line 69: same

neSpecc (changes requested):

src/application/services/useNoteSettings.ts
```diff
@@ -65,14 +65,18 @@ export default function (): UseNoteSettingsComposableState {
...
    *
    * @param id - Note id
-   * @param data - Note settings data with new values
+   * @param newIsPublicValue - new isPublic
    */
-  const update = async (id: NoteId, data: Partial<NoteSettings>): Promise<void> => {
-    noteSettings.value = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
-  };
+  async function updateIsPublic(id: NoteId, newIsPublicValue: boolean): Promise<void> {
+    const response = await noteSettingsService.patchNoteSettingsByNoteId(id, { isPublic: newIsPublicValue });
```
neSpecc, line 74: `response` is a transport-related term, it should not be used in this context.

src/application/services/useNoteSettings.ts
```diff
@@ -65,14 +65,18 @@ export default function (): UseNoteSettingsComposableState {
...
+
+    if (noteSettings.value) {
```
neSpecc, line 76: this case is not clear, add a jsdoc dscribing why do you need it. When it can be null?

src/presentation/pages/NoteSettings.vue
```diff
@@ -64,23 +56,21 @@ const invitationLink = computed(
...
 /**
- * Regenerate invitation hash
+ * Change isPublic property
  */
-async function regenerateHash() {
-  revokeHash(props.id);
+async function changeAccess() {
+  const isPublic = noteSettings.value?.isPublic;
+
+  if (isPublic) {
```
neSpecc, line 71: unclear statement, it won't allow setting isPublic=false.

neSpecc (changes requested):

src/application/services/useNoteSettings.ts
elizachi: You mean I should rename it?

neSpecc: service returns the note settings, you can pick the new value using desctructor

```ts
const { isPublic } = await ...
```

src/application/services/useNoteSettings.ts
```diff
@@ -65,14 +65,21 @@ export default function (): UseNoteSettingsComposableState {
...
+   * @param newIsPublicValue - new isPublic
    */
-  const update = async (id: NoteId, data: Partial<NoteSettings>): Promise<void> => {
-    noteSettings.value = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
-  };
+  async function updateIsPublic(id: NoteId, newIsPublicValue: boolean): Promise<void> {
+    const updatedIsPublic = await noteSettingsService.patchNoteSettingsByNoteId(id, { isPublic: newIsPublicValue });
+
+    /**
+     * Check if note setting is not empty
```
neSpecc, line 77: it still seems like a crutch around TS issue

src/presentation/pages/NoteSettings.vue
```diff
@@ -64,23 +56,21 @@ const invitationLink = computed(
...
 /**
- * Regenerate invitation hash
+ * Change isPublic property
  */
-async function regenerateHash() {
-  revokeHash(props.id);
+async function changeAccess() {
+  const isPublic = noteSettings.value?.isPublic;
+
+  if (isPublic !== undefined) {
```
neSpecc, line 71: if user changes access, it means that note settings is defined. So this case has no sense

elizachi: But we still should to check noteSettongs.value, in other case we have error

neSpecc: I'd suggest using `!` instead, since we know that it can't be undefined here.

src/application/services/useNoteSettings.ts
elizachi: in general, notes are updated without explicitly assigning the result to noteSettings.value, but i think this option is more correct

src/presentation/pages/NoteSettings.vue
elizachi: Okay

src/application/services/useNoteSettings.ts
elizachi: Or I can do like in revokeHash function, smth like:
```
/**
     * Check if note setting is not empty
     */
    if (noteSettings.value) {
      noteSettings.value = { ...noteSettings.value, isPublic };
    }
    ```

Comment (neSpecc) [reply src/application/services/useNoteSettings.ts:77]: let's leave it as is

Comment (neSpecc) [inline src/application/services/useNoteSettings.ts:77]: Bad jsdoc — it duplicates a logic. It should describe when noteSettings can be undefined instead

File: src/application/services/useNoteSettings.ts [hunk 12]
```diff
@@ -65,14 +65,21 @@ export default function (): UseNoteSettingsComposableState {
...
+   * @param newIsPublicValue - new isPublic
    */
-  const update = async (id: NoteId, data: Partial<NoteSettings>): Promise<void> => {
-    noteSettings.value = await noteSettingsService.patchNoteSettingsByNoteId(id, data);
-  };
+  async function updateIsPublic(id: NoteId, newIsPublicValue: boolean): Promise<void> {
+    const { isPublic } = await noteSettingsService.patchNoteSettingsByNoteId(id, { isPublic: newIsPublicValue });
+
+    /**
+     * Check in case noteSettings has not yet been created and is set to null
```
neSpecc, line 77: > noteSettings has not yet been created

it's impossible. Note Settings are being created along with note.

elizachi: Then maybe just Check in case noteSettings is set to null?

neSpecc: I'm suggesting to leave the code as is, but improve docs describing a real-world case

neSpecc, line 77:
```suggestion
     * If note settings were not loaded till this moment for some reason, do nothing
```

neSpecc (approved):

PR #154, opened 2024-03-30
feat(codex-ui): Implement icon element

GoldenJaden:
image


TatianaFomina:
there are some conflicts in the PR 😢

codex-ui/dev/Playground.vue
```diff
@@ -65,14 +65,19 @@
       size="large"
     />
 
+    <Heading :level="3"> Icon </Heading>
+
+    <Icon :src="IconChecklist"/> <Icon :src="IconLoader"/>
```
TatianaFomina, line 100: `IconChecklist` is not icon source. It's text containing icon html

TatianaFomina: Maybe incapsulate all the icon data imports inside `Icon` component? It would be easier if `Icon` had a prop `name` where you could specify icon name, in this case 'checkbox'. And the `Icon` component itself would contain all the icons imports and mapping between names and data

TatianaFomina:
There are some conflicts ☠️
image

elizachi (approved):

codex-ui/dev/Playground.vue
GoldenJaden: Done

neSpecc (approved):

PR #185, opened 2024-04-24
feat(note-settings): Implement switch component to the note settings

elizachi:
Now we have switch component instead checkbox
image

### DIFF
--- a/src/presentation/pages/NoteSettings.vue
+++ b/src/presentation/pages/NoteSettings.vue
@@ -16,7 +16,6 @@
         <template #right>
           <Switch
             v-model="isPublic"
-            :value="noteSettings.isPublic"
             @click="changeAccess"
           />
         </template>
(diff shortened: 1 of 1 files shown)


e11sy (approved):

src/presentation/pages/NoteSettings.vue
```diff
@@ -14,10 +14,10 @@
     >
       <Row :title="t('noteSettings.publish')">
         <template #right>
-          <Checkbox
-            v-model:checked="noteSettings.isPublic"
-            label=""
-            @update:checked="changeAccess"
+          <Switch
+            v-model="isPublic"
+            :value="noteSettings.isPublic"
```
neSpecc, line 19: why do we need to pass both `value` and `v-model`?  

v-model includes value

src/presentation/pages/NoteSettings.vue
```diff
@@ -72,11 +72,13 @@ async function regenerateHash() {
   revokeHash(props.id);
 }
 
+const isPublic = ref(noteSettings.value?.isPublic);
```
neSpecc, line 75: jsdoc are missed

neSpecc (approved):

PR #190, opened 2024-04-26
noteTools added for POST and PATCH /note

e11sy:
## Problem
```
noteTools: {
  name: string,
  id: string
}
```
now noteTools are not send to api on POST and PATCH /note, but they are required in body, that is why API returns 500

## Solution
- added noteTools entity and noteTools formalizatoin for useNote service
- added noteTools, that are formed from content blocks ( via user tools )
- now on POST and PATCH /note `tools: noteTools` are included to body

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -179,7 +179,6 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     }
 
     const usedNoteTools = content.blocks.map((block) => {
-
       const blockTool = (tools.value as EditorTool[]).find((tool) => tool.name === block.type)
 
       /**
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -4,6 +4,7 @@ import type { Note, NoteContent, NoteId } from '@/domain/entities/Note';
 import { useRouter } from 'vue-router';
 import type { NoteDraft } from '@/domain/entities/NoteDraft';
 import type EditorTool from '@/domain/entities/EditorTool';
+import { useAppState } from './useAppState';
 
 /**
  * Creates base structure for the empty note:
@@ -164,9 +165,10 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    *
    * @param content - Note content (Editor.js data)
    * @param parentId - Id of the parent note. If null, then it's a root note
-   * @param tools
    */
-  async function save(content: NoteContent, parentId: NoteId | undefined, tools: EditorTool[]): Promise<void> {
+  async function save(content: NoteContent, parentId: NoteId | undefined): Promise<void> {
+    const { userEditorTools } = useAppState();
+
     if (note.value === null) {
       throw new Error('Note is not loaded yet');
     }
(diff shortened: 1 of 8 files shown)


src/application/services/useNote.ts
```diff
@@ -166,15 +167,24 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+    const { userEditorTools } = useAppState();
+
     if (note.value === null) {
       throw new Error('Note is not loaded yet');
     }
 
+    /**
+     * Format specified tools
+     */
+    const specifiedNoteTools = content.blocks.map((block) => {
```
elizachi, line 179: What means specified note tools?

elizachi (approved):

src/application/services/useNote.ts
```diff
@@ -166,15 +167,24 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+      return { name: block.type, id: (userEditorTools.value ?? []).find((tool) => tool.name === block.type)!.id };
```
neSpecc, line 180: lets make the `resolveToolByName()` method

src/domain/entities/Note.ts
```diff
@@ -10,6 +11,18 @@ export type NoteId = string;
  */
 export type NoteContent = OutputData;
 
+export type NoteTools = {
```
neSpecc, line 14: add entity description

src/application/services/useNote.ts
TatianaFomina, line 180: note can contain tools that are not installed by the user. For example, there are 2 collaborators on the note. One has list plugin installed, the other does not. If the second user tries to update the note, what will happen?

src/application/services/useNote.ts
e11sy: these are tools, that would be specified in the api request

src/application/services/useNote.ts
```diff
@@ -159,6 +166,19 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+  /**
+   * Returns list of tools used in the note
+   *
+   * @param content - content of the note
+   */
+  async function resolveToolsByContent(content: NoteContent): Promise<NoteTool[]> {
+    const { allTools } = useTools(noteTools);
+
+    return content.blocks.map((block) => {
+      return { name: block.type, id: (allTools.value ?? []).find((tool) => tool.name === block.type)!.id };
```
neSpecc, line 178: its hard to read, please, write it via the separate statements

src/application/services/useNote.ts
```diff
@@ -159,6 +166,19 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
 
+  /**
+   * Returns list of tools used in the note
+   *
+   * @param content - content of the note
+   */
+  async function resolveToolsByContent(content: NoteContent): Promise<NoteTool[]> {
+    const { allTools } = useTools(noteTools);
+
+    return content.blocks.map((block) => {
```
neSpecc, line 177: map won't be unique

src/application/services/useTools.ts
```diff
@@ -63,20 +65,21 @@ export function useTools(noteTools: Ref<EditorTool[]>): {
...
       }
-      tools.value = await downloadTools(allTools);
+      tools.value = await downloadTools(allTools.value);
     },
     { immediate: true }
   );
 
   return {
     tools,
+    allTools,
```
neSpecc, line 83: There is a problem with naming: 
seems like `tools` and `allTools` contains the same set. But first contains downloaded something, and all tools contains just info. 

I'w suggest:

`tools` -> `toolsConnected`
`allTools` -> `tools`

src/application/services/useNote.ts
TatianaFomina: ➕

TatianaFomina (approved):

PR #199, opened 2024-05-16
feat[eslint]: updated eslint and codex-eslint-config to latest

e11sy:
- updated dependencies
- fixed some of the eslint problems

Note that all eslint errors and warnings would be solved in further development process


src/application/services/useNote.ts
```diff
@@ -190,7 +187,8 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     }
 
     await noteService.updateNoteContent(currentId.value, content);
-    note.value = { ...note.value, content };
+    note.value = { ...note.value,
+      content };
```
neSpecc, line 191: bad formatting. It should format such cases as 

```suggestion
    note.value = { 
      ...note.value,
      content 
    };
```

neSpecc: we can create an issue in the Config for that

neSpecc (approved):

elizachi (approved):

src/application/services/useNote.ts
e11sy: done

PR #208, opened 2024-06-16
Eslint: add no unsafe assignment rule

kseniaSs:
Add eslint's rule "no-unsafe-assignment" to project


e11sy:
seems like it never would be done

PR #209, opened 2024-06-19
refactor(codex-ui): support input leading icon

elizachi:
The input component now supports the icon on the left
An icon has been added to the menu context

image
image


e11sy (approved):

codex-ui/src/vue/components/input/Input.vue
```diff
@@ -41,14 +54,28 @@ onMounted(() => {
 });
 </script>
 
-<style lang="postcss" module>
+<style module>
 .input {
-  background-color: var(--base--bg-secondary);
+  display: flex;
   border: 0;
   outline: 0;
```
neSpecc, line 61: div has no border and outline, so these lines has no sense

codex-ui/src/vue/components/input/Input.vue
```diff
@@ -41,14 +54,28 @@ onMounted(() => {
...
-  color: var(--base--text);
-  font-size: inherit;
-  font-family: inherit;
+  background-color: var(--base--bg-secondary);
+  gap: var(--v-padding);
+  align-items: center;
+
+  &__wrapper {
+    flex: 1;
+    background-color: var(--base--bg-secondary);
```
neSpecc, line 68: you need to decide, where background should be.

codex-ui/src/vue/components/input/Input.vue
elizachi: There is a small problem: if we remove `border: 0;`, a border will appear around the parent container, so I suggest removing this line

image

kloV148 (approved):

codex-ui/src/vue/components/menu/ContextMenu.vue
```diff
@@ -38,8 +41,16 @@ withDefaults(
      * Array of items for context menu
      */
     items: ContextMenuItems[];
+
+    /**
+     * Icon displayed in ipnut
+     */
+    searchIcon?: string;
```
neSpecc, line 48: prop is redundant

neSpecc (commented):
- clickable zone is too small. Input should be focused in any area  inside.
- size is larger than in Figma (38 vs 36)

codex-ui/src/vue/components/input/Input.vue
```diff
@@ -1,14 +1,21 @@
...
-    :disabled="props.disabled"
-  >
+  <div :class="[$style['input'], `${$style.input}--${size}`]">
+    <Icon
+      v-if="icon"
+      :name="icon"
+    />
+    <input
+      v-model="model"
+      :class="[$style['input__wrapper'], 'text-ui-base']"
```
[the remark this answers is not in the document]
elizachi: I can rename to smth like "style"... Do you have any ideas?

elizachi (commented):
@neSpecc I can also add placeholder text in this request

codex-ui/src/vue/components/input/Input.vue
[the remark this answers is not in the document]
neSpecc: all elements are styled. It's better to call it with semantic value

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -58,6 +61,7 @@ withDefaults(
 
   &__input {
     padding: var(--v-padding) var(--h-padding);
+    color: var(--base--text-secondary)
```
neSpecc, line 64: it seems redundant, placeholder is expected instead

codex-ui/src/vue/components/input/Input.vue
```diff
@@ -1,14 +1,25 @@
...
+    @click="focusInput"
   >
+    <Icon
+      v-if="icon"
+      :name="icon"
+    />
+    <input
+      ref="textInput"
+      v-model="model"
+      :class="[$style['input__semantic'], 'text-ui-base']"
```
neSpecc, line 13:
```suggestion
      :class="[$style['input__editable-zone'], 'text-ui-base']"
```

codex-ui/src/vue/components/input/Input.vue
```diff
@@ -26,29 +37,60 @@ const props = withDefaults(
...
   model.value = props.value;
 });
 </script>
 
-<style lang="postcss" module>
+<style module>
 .input {
-  background-color: var(--base--bg-secondary);
+  display: flex;
   border: 0;
```
neSpecc, line 72: border is useless, since div has no border itself

codex-ui/src/vue/components/context-menu/ContextMenu.vue
elizachi: you mean move Icon in placeholder?

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -4,7 +4,11 @@
       v-if="showSearch"
       :class="$style['context-menu__search']"
     >
-      <Input :class="$style['context-menu__input']" />
+      <Input
+        :class="$style['context-menu__input']"
```
neSpecc, line 8: this class seems redundant

codex-ui/src/vue/components/input/Input.vue
```diff
@@ -1,14 +1,27 @@
...
-    v-model="model"
-    :class="[$style.input, `${$style.input}--${size}`, 'text-ui-base']"
-    :disabled="props.disabled"
+  <div
+    :class="[$style['input'], `${$style.input}--${size}`]"
+    @click="focusInput"
   >
+    <Icon
+      v-if="icon"
+      :style="{ color: 'var(--base--text-secondary)' }"
```
neSpecc, line 8: use css instead

neSpecc (approved):

PR #218, opened 2024-07-01
feat: add simple add tool design

GeekaN2:
20 cents design

image


src/application/i18n/messages/en.json
neSpecc, line 1: wrong placeholders

slaveeks (approved):

neSpecc (approved):

PR #220, opened 2024-07-01
feat(note settings page): setup page base layout

kloV148:
The first part of the implementation of the new design for the note settings page

Added:
- base layout
- Header
- Subheading is calculated from the content of the note

Снимок экрана 2024-07-01 в 20 46 19


src/application/services/useNoteSettings.ts
```diff
@@ -87,8 +94,14 @@ export default function (): UseNoteSettingsComposableState {
   const load = async (id: NoteId): Promise<void> => {
     noteSettings.value = await noteSettingsService.getNoteSettingsById(id);
     const response = await noteService.getNoteById(id);
+    const firstNoteBlock = response.note.content.blocks[0];
+    const limitCharsForNoteTitle = 50;
 
     parentNote.value = response.parentNote;
+
+    if (firstNoteBlock !== null && Boolean(firstNoteBlock.data.text)) {
```
elizachi, line 102:
```suggestion
    if (firstNoteBlock && firstNoteBlock.data.text) {
```

src/presentation/pages/NoteSettings.vue
```diff
@@ -1,53 +1,69 @@
...
-    >
-      <Row :title="t('noteSettings.publish')">
-        <template #right>
-          <Switch
-            v-model="isPublic"
-            @click="changeAccess"
+  <div
+    v-if="noteSettings"
+  >
+    <div class="note-settings">
```
elizachi, line 5: You can remove this div and add v-if to the next

src/application/services/useNoteSettings.ts
neSpecc: you can use the `hasValue()` util

kloV148: I cant do like this, because of eslint warnings

```
  105:9   error  Unexpected object value in conditional. The condition is always true                  @typescript-eslint/strict-boolean-expressions
  105:27  error  Unexpected any value in conditional. An explicit comparison or type cast is required  @typescript-eslint/strict-boolean-expressions
  ```

File: src/application/services/useNoteSettings.ts [hunk 3]
```diff
@@ -164,6 +177,7 @@ export default function (): UseNoteSettingsComposableState {
 
   return {
     parentNote,
+    noteTitle,
```
neSpecc, line 180: its is not a part of the note settings

src/presentation/pages/NoteSettings.vue
```diff
@@ -1,53 +1,69 @@
...
+          :caption="t('noteSettings.parentNoteCaption')"
+          :disabled="parentNote != undefined"
+          :placeholder="t('noteSettings.parentNotePlaceholder')"
+          @input="setParentDebounced"
+        />
+        <Section
+          :title="t('noteSettings.availabilityTitle')"
+          :caption="t('noteSettings.availabilityCaption')"
+        >
+          <Row :title="t('noteSettings.publish')">
```
neSpecc, line 35: there should be "Note is published"

src/presentation/pages/NoteSettings.vue
```diff
@@ -1,53 +1,69 @@
...
+          {{ noteTitle }}
+        </Heading>
+      </div>
+      <div class="form">
+        <Field
+          v-model="parentURL"
+          :title="t('noteSettings.parentNote')"
+          size="medium"
+          :caption="t('noteSettings.parentNoteCaption')"
+          :disabled="parentNote != undefined"
```
e11sy, line 27: !==

neSpecc (commented):
- name pr using our naming convention https://github.com/vuejs/core/blob/main/.github/commit-convention.md
- update PR description: right now it's not clear why these changes are different from Figma. Is it the first part of the implementation?

src/presentation/pages/NoteSettings.vue
```diff
@@ -153,16 +169,39 @@ onMounted(async () => {
...
-<style scoped>
-.control__button {
-  padding: var(--spacing-xxs) var(--spacing-ms);
-  align-items: center;
+<style setup lang="postcss" scoped>
+@import '@/presentation/styles/typography.pcss';
+
+h1 {
+  @apply --text-heading-1;
+}
```
e11sy, line 180: seems redundant

src/application/services/useNoteSettings.ts
```diff
@@ -20,6 +20,11 @@ interface UseNoteSettingsComposableState {
    */
   parentNote: Ref<Note | undefined>;
 
+  /**
+   * Title of the note
+   */
+  noteTitle: Ref<string>;
+
```
e11sy, line 27: note can be untitled (e.g. note contains only image), we need default value like `Untitled`

neSpecc (approved):

PR #231, opened 2024-07-02
feat(codex-ui): controls dimensions

neSpecc:
Now the "Controls Dimensions" concept is implement.

It is the way to change elements paddings and radius to "small", "medium" or "large"

## How to use

Just add the `data-dimensions` attribute to any element. All the subtree will inherit new dimensions.

## Additional changes

- `Button.vue` 
  - `size` property removed. 
  - `style` property refactored since it was being conflicted with the native "style" property — use separate `secondary`, `descrtuctive` props instead
  - trailing icon support added
  - standalone icon support added
- `Section.vue` - `size` property removed. 
- `Field.vue` - `size` property removed.  
- `Input.vue` - `size` property removed. 
- `Row.vue` - `subtle` property renamed to `subtitle`. 

image


elizachi (approved):

codex-ui/src/vue/components/button/Button.vue
```diff
@@ -1,32 +1,42 @@
 <template>
   <button
-    :class="[$style.button, `${$style.button}--${size}`, `${$style.button}--${style}`, 'text-ui-base-medium', icon && `${$style.button}--with-icon`]"
+    :class="[
+      $style.button,
+      `${$style.button}--${style}`,
```
e11sy, line 5: redundant if style prop is deprecated

codex-ui/src/vue/components/button/Button.vue
```diff
@@ -1,32 +1,42 @@
...
      */
-    size?: ButtonSize;
+    secondary?: boolean;
 
     /**
-     * The style of the button
+     * Pass this attribue for negative actions
      */
-    style?: ButtonStyle;
+    destructive?: boolean;
```
e11sy, line 39: maybe we can just rename style to smth else to avoid conflict? 
now if we specify both secondary and destructive will be applied only secondary style

codex-ui/src/vue/components/button/Button.vue
neSpecc: style is a computed property now

codex-ui/src/vue/components/button/Button.vue
neSpecc: I think it's still more elegant way to write

```vue
<Button secondary />
```

instead of 

```vue
<Button style="secondary" />
```

and it is similar with native attributes of the [HTMLButtonElement](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button)

PR #235, opened 2024-07-02
refactor(codex-ui):search in context menu

elizachi:
Now it is possible to search for an item among menu context items.
The items of show-case array has been modified to test the `context menu`.

image

Added a filtering feature that allows you to search among the `menu context` items. If the entered string is a substring of an item, it appears in the menu context. Separators remain between elements, but are removed from the beginning and end of the derived array.

image

If no item is found, the menu context displays the corresponding message.

image

### DIFF
--- a/codex-ui/dev/pages/components/ContextMenu.vue
+++ b/codex-ui/dev/pages/components/ContextMenu.vue
@@ -9,7 +9,7 @@
   <ContextMenu
     :show-search="true"
     :items="contextMenuItems"
-    :message="'Noting found'"
+    :message="'Nothing found'"
   />
 </template>
 
--- a/codex-ui/dev/pages/components/ContextMenu.vue
+++ b/codex-ui/dev/pages/components/ContextMenu.vue
@@ -43,6 +43,9 @@ const contextMenuItems: ContextMenuItem[] = [
     onActivate: console.log,
   },
   {
+    type: 'separator',
+  },
+  {
     type: 'default',
     title: 'Text',
     icon: 'Text',
--- a/codex-ui/dev/pages/components/ContextMenu.vue
+++ b/codex-ui/dev/pages/components/ContextMenu.vue
@@ -9,7 +9,7 @@
   <ContextMenu
     :show-search="true"
     :items="contextMenuItems"
-    :message="'Nothing found'"
+    message="Nothing found"
   />
 </template>
 
--- a/codex-ui/dev/pages/components/ContextMenu.vue
+++ b/codex-ui/dev/pages/components/ContextMenu.vue
@@ -9,6 +9,7 @@
   <ContextMenu
     :show-search="true"
     :items="contextMenuItems"
+    :message="'Noting found'"
   />
 </template>
 
--- a/codex-ui/dev/pages/components/ContextMenu.vue
+++ b/codex-ui/dev/pages/components/ContextMenu.vue
@@ -9,7 +9,6 @@
   <ContextMenu
     :show-search="true"
     :items="contextMenuItems"
-    message="Nothing found"
   />
 </template>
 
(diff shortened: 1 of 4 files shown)


codex-ui/dev/pages/components/ContextMenu.vue
```diff
@@ -9,6 +9,7 @@
   <ContextMenu
     :show-search="true"
     :items="contextMenuItems"
+    :message="'Nothing found'"
```
e11sy, line 12: use i18n

slaveeks, line 12:
```suggestion
    message="Nothing found"
```

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -41,10 +48,56 @@ withDefaults(
      * Array of items for context menu
      */
     items: Item[];
+
+    /**
+     * Message what occurs as a result of the search
+     */
+    message: string;
```
e11sy, line 55: i think that is we use it just for search, it should be moved to constants as "Nothing found" or smth same

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -41,10 +48,56 @@ withDefaults(
...
+    }
+  });
+
+  if (searchedItems.length > 0) {
+    return searchedItems[0].type === 'separator'
+      ? searchedItems.slice(1)
+      : searchedItems;
+  }
+
+  return searchedItems;
```
slaveeks, line 90: add some docs to this logic please

codex-ui/dev/pages/components/ContextMenu.vue
neSpecc, line 12: unclear property name

neSpecc: I think we can get rid of it at all since it is a static text, you can pass it inside the ContextMenu to item

e11sy:
image
last separator should be removed also

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -41,10 +48,56 @@ withDefaults(
...
+ */
+const filteredItems = computed(() => {
+  if (searchTerm.value === '') {
+    return props.items;
+  }
+
+  const searchedItems = props.items.filter((item) => {
+    if (item.type === 'message') {
+      return false;
+    } else if (item.type === 'separator') {
```
e11sy, line 76: remove this and check (*) comment

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -5,22 +5,29 @@
...
+      <template v-if="filteredItems.length > 0">
+        <ContextMenuItem
+          v-for="(item, index) in filteredItems"
+          :key="index"
+          :item="item"
+        />
+      </template>
+      <template v-else>
+        <ContextMenuItem :item="messageItem" />
+      </template>
```
e11sy, line 24: (*) comment:
if we do not filter items with type: message, we will have two cases
- items are only with type message element
```
{ type: 'message'}
``` 
- some items without type message

codex-ui/src/vue/components/context-menu/ContextMenu.vue
e11sy, line 90: (*) we have this case only when `serchedItems.lenght === 0`, so we can just return message item here

e11sy (changes requested):

codex-ui/src/vue/components/context-menu/ContextMenu.vue
```diff
@@ -45,6 +58,76 @@ withDefaults(
...
+ * @param items - filtered items array
+ * @returns { Item[] } - array without unnecessary separators
+ */
+function removeSeparators(items: Item[]): Item[] {
+  if (items.length > 0) {
+    if (items[0].type === 'separator') {
+      items = items.slice(1);
+    }
+    if (items[items.length - 1].type === 'separator') {
+      items = items.slice(0, items.length - 1);
```
e11sy, line 116: this is pop()

e11sy (approved):

kloV148 (approved):

PR #237, opened 2024-07-03
feat(note page): note header implementation

elizachi:
Note header added

NoteHeader hide when note **not created** yet

image

NoteHeader

image


codex-ui/dev/Playground.vue
neSpecc, line 1: it is not a component of the CodeX UI. It's better to implement it inside NoteX

codex-ui/src/vue/components/note-header/NoteHeader.vue
```diff
@@ -0,0 +1,83 @@
+<template>
+  <div
+    :class="$style['note-header']"
+    :style="{ '--opacity': props.opacity }"
```
neSpecc, line 4: you can manage visibility of the header using `v-show` outside

codex-ui/src/vue/components/note-header/NoteHeader.vue
```diff
@@ -0,0 +1,83 @@
...
+<script setup lang="ts">
+import { computed } from 'vue';
+import Button from '../button/Button.vue';
+
+const props = withDefaults(
+  defineProps<{
+    /**
+     * Explanatory caption for the date
+     */
+    description?: string;
```
neSpecc, line 31: is it used?

codex-ui/src/vue/components/note-header/NoteHeader.vue
```diff
@@ -0,0 +1,83 @@
...
+  width: 100%;
+  display: flex;
+  height: min-content;
+  opacity: var(--opacity);
+  justify-content: space-between;
+  padding: var(--spacing-s) var(--spacing-m);
+
+  &__left {
+    display: flex;
+    align-items: center;
```
neSpecc, line 63: redundant styles

codex-ui/src/vue/components/note-header/NoteHeader.vue
```diff
@@ -0,0 +1,83 @@
...
+
+  &__button {
+    padding: var(--spacing-s);
+    color: var(--base--text-secondary);
+    background-color: var(--base--bg-secondary);
+
+    &:hover {
+      background-color: var(--base--bg-secondary-hover);
+      color: var(--base--text-secondary-hover);
+    }
```
neSpecc, line 80: redundant styles

src/presentation/pages/Note.vue
```diff
@@ -1,4 +1,9 @@
 <template>
+  <NoteHeader
+    :opacity="note ? 1 : 0"
+    description="Last edit"
```
neSpecc, line 4: redundant property

src/presentation/pages/Note.vue
```diff
@@ -136,6 +143,20 @@ function unlinkButton(): void {
...
+ *
+ * @param newNote - Note entity
+ * @returns {string | undefined}
+ */
+function getLastUpdate(newNote: Note): string | undefined {
+  if (newNote.updatedAt === undefined) {
+    return '';
+  }
+
+  return formatShortDate(newNote.updatedAt);
```
neSpecc, line 157: place date formatting inside the Note Header

codex-ui/src/vue/components/note-header/NoteHeader.vue
elizachi: The v-show hides the component at all, once hidden it just disappears. Figma says it should stay, just be transparent.
image

`Using v-show`
image

`Using opacity`
image

codex-ui/src/vue/components/note-header/NoteHeader.vue
elizachi: Yes, here
image

I can remove it and left just static text

src/presentation/components/note-header/NoteHeader.vue
```diff
@@ -0,0 +1,114 @@
...
+  if (props.id === null) {
+    throw new Error('Note is Empty');
+  }
+  router.push(`/note/${props.id}/new`);
+}
+
+/**
+ * Move to note settings
+ */
+function getNoteSettings(): void {
```
e11sy, line 74: maybe
```
function goToNoteSettings(): void {
```
or
```
function redirectToNoteSettings(): void {
```
?

src/presentation/components/note-header/NoteHeader.vue
```diff
@@ -0,0 +1,114 @@
...
+  if (props.id === null) {
+    throw new Error('Note is Empty');
+  }
```
e11sy, line 77: maybe add this condition to template `settingsButton` part?

src/presentation/pages/Note.vue
```diff
@@ -1,43 +1,39 @@
...
-      ref="editor"
-      :tools="loadedTools"
-      :content="note.content"
-      :read-only="!canEdit"
-      @change="noteChanged"
+  <div>
+    <NoteHeader
+      :id="props.id"
+      :opacity="props.id ? 1 : 0"
     />
```
e11sy, line 6: i think we need to move it into v-else
should we really see noteHeader, when note is loading?

src/presentation/pages/Note.vue
```diff
@@ -1,24 +1,38 @@
...
-        v-if="props.id != null"
-        @click="createChildNote"
-      >
-        {{ t('note.createChildNote') }}
-      </Button>
+  <div>
+    <NoteHeader
+      :id="props.id"
+      :style="{ '--opacity': props.id ? 1 : 0 }"
+    >
```
e11sy, line 6:
```suggestion
    <NoteHeader
      :id="id"
      :style="{ '--opacity': id ? 1 : 0 }"
    >
```

src/presentation/pages/Note.vue
```diff
@@ -1,22 +1,36 @@
...
-    <div>
-      <Button
-        v-if="props.id != null"
-        @click="createChildNote"
-      >
-        {{ t('note.createChildNote') }}
-      </Button>
+  <div>
+    <NoteHeader
+      :id="id"
```
neSpecc, line 4: why id is used?

src/presentation/pages/Note.vue
```diff
@@ -1,22 +1,36 @@
...
+      :style="{ '--opacity': id ? 1 : 0 }"
```
neSpecc, line 5: right now it is shown event when not is not loaded. Let's show it only when note is loaded?

src/presentation/pages/Note.vue
```diff
@@ -54,6 +70,41 @@ const { note, noteTools, save, noteTitle, canEdit } = useNote({
...
+ */
+watch(note, () => {
+  if (note.value !== null && 'updatedAt' in note.value) {
+    const updatedAt = note.value.updatedAt;
+
+    if (typeof updatedAt === 'string') {
+      lastEdit.value = 'Last edit ' + formatShortDate(updatedAt);
+    }
+  }
+});
```
neSpecc, line 106: Can we get rid of `lastEdit` variable and use `'Last edit ' + formatShortDate(note.value.updatedAt);` directly in template since note is reactive

src/presentation/pages/Note.vue
neSpecc, line 5: lets add small transition via css

e11sy (approved):

src/presentation/pages/Note.vue
```diff
@@ -54,6 +70,41 @@ const { note, noteTools, save, noteTitle, canEdit } = useNote({
   id: noteId,
 });
 
+const lastEdit = ref<string>('Last edit ');
```
kloV148, line 73: Maybe display something like "Last edit just now"? It looks strange now

src/presentation/pages/Note.vue
```diff
@@ -54,6 +70,41 @@ const { note, noteTools, save, noteTitle, canEdit } = useNote({
...
+
+/**
+ * Loads the time when the note was last modified when opened
+ */
+watch(note, () => {
+  if (note.value !== null && 'updatedAt' in note.value) {
+    const updatedAt = note.value.updatedAt;
+
+    if (typeof updatedAt === 'string') {
+      lastEdit.value = 'Last edit ' + formatShortDate(updatedAt);
```
kloV148, line 103: Can we use i18n here? Or we dont have this ability rn

src/presentation/pages/Note.vue
elizachi: I can write `a few seconds ago` like in our finction

src/presentation/pages/Note.vue
elizachi: I think we can, i'll do it

src/presentation/pages/Note.vue
kloV148: Sounds good

src/presentation/pages/Note.vue
elizachi: In this case we have propblem with types
image

I think it's best to leave it as it is

src/presentation/pages/Note.vue
elizachi: But then we have "jumping" block of Editor

Without NoteHeader
image

With NoteReader
image

We have 52 pixels difference

elizachi: If it's not critical I can use v-show to hide it when note is Draft

neSpecc: it won't jump because you're hiding via opacity. Just hight it when note is not loaded as well

src/application/i18n/messages/en.json
```diff
@@ -79,7 +79,9 @@
     "createChildNote": "Create child note",
     "editor": {
       "placeholder": "Write something or press / to select a tool"
-    }
+    },
+    "lastEdit": "Last edit",
+    "defaultLastEdit": "Last edit a few seconds ago"
```
neSpecc, line 84: seems redundant

src/presentation/pages/Note.vue
neSpecc: because you don't need to use `.value` in template

src/presentation/pages/Note.vue
```diff
@@ -1,22 +1,35 @@
...
+        <Button
+          secondary
+          icon="Plus"
+          @click="createChildNote"
+        />
+        <Button
+          secondary
+          icon="EtcHorisontal"
+          @click="redirectToNoteSettings"
+        />
```
neSpecc, line 19: these buttons should be only visible when you have a write access

src/presentation/pages/Note.vue
elizachi: image

The result same without .value

neSpecc (approved):

PR #240, opened 2024-07-03
feat(note-settings)/parent-note-card

kloV148:
From now we can see parent note card in note settings page.
Also unlink button was removed from note page. Now you can unlink note from not settings page. 
ParentNote and setParent were moved to useNote

**With parent note**
Снимок экрана 2024-07-03 в 17 18 59

**Without parent note**
Снимок экрана 2024-07-03 в 17 34 00

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -289,7 +289,6 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
       if (error instanceof Error) {
         window.alert(error.message);
       }
-      throw error;
     }
   };
 
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -71,13 +71,6 @@ interface UseNoteComposableState {
   unlinkParent: () => Promise<void>;
 
   /**
-   * Set parent for the note
-   * @param id - Child note id
-   * @param newParentURL - New parent note URL
-   */
-  setParent: (id: NoteId, newParentURL: string) => Promise<void>;
-
-  /**
    * Defines if user can edit note
    */
   canEdit: Ref<boolean>;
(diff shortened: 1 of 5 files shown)


src/presentation/pages/NoteSettings.vue
```diff
@@ -210,4 +250,16 @@ onMounted(async () => {
...
+.change-parent{
+  display: flex;
+  flex-direction: column;
+  gap: var(--v-padding);
+
+  &__button{
+    display: flex;
+    justify-content: flex-end;
+    flex-grow: 1;
+  }
```
e11sy, line 263: i think card slot alignment was fixed in some of previous pr's, try to update from main and remove this style

kloV148: Yes, it was fixed. This style was removed in last commit

e11sy (approved):

src/application/services/useNote.ts
```diff
@@ -270,6 +277,22 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     parentNote.value = undefined;
   }
 
+  /**
+   * Set parent for the note
+   * @param id - Child note id
+   * @param newParentURL - New parent note URL
+   */
+  const setParent = async (id: NoteId, newParentURL: string): Promise<void> => {
```
neSpecc, line 285: use named function instead of anonymous, it's better for debugging
```suggestion
  async function setParent(id: NoteId, newParentURL: string): Promise<void> {
```

neSpecc: I'd suggest leaving `setParent` at the `useNoteSettigns` service because we can change parent only from the settings page

src/application/services/useNote.ts
```diff
@@ -270,6 +277,22 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+    try {
+      parentNote.value = await noteService.setParentByUrl(id, newParentURL);
+    } catch (error) {
+      if (error instanceof Error) {
+        window.alert(error.message);
+      }
+      throw error;
```
neSpecc, line 292: redundant throw

neSpecc (approved):

PR #251, opened 2024-07-05
feat(auth): Implemented authRequired decoration for routes 

e11sy:
image


src/application/router/routes.ts
```diff
@@ -127,17 +135,18 @@ const routes: RouteRecordRaw[] = [
...
     props: route => ({
-      invitationHash: String(route.params.hash),
+      redirect: String(route.query.redirect),
     }),
     meta: {
-      pageTitleI18n: 'pages.joinTeam',
+      pageTitleI18n: 'pages.authorization',
       discardTabOnLeave: true,
     },
   },
```
elizachi, line 152: Why remove join?

e11sy: this was duplicated join route, i replaced duplicated join with authorization 
image

elizachi (approved):

src/application/services/useAuthRequired.ts
```diff
@@ -0,0 +1,40 @@
...
+  const { user } = useAppState();
+
+  /**
+   * Check if user is authorized
+   * If authorization is in process we treat user as unauthorized
+   * When oauth will work, it will be treated as he authorized manually
+   * @returns true if user is authorized, false otherwise
+   */
+  function isUserAuthorized(): boolean {
+    return (user.value !== null || user.value === undefined);
```
neSpecc, line 22: `user.value === undefined` — it is not authorized

e11sy: fixed

neSpecc (approved):

PR #253, opened 2024-07-05
chore(home): Home page styles updated

e11sy:
Now home page is implemented
Now load next page will append next page notes to the currnet notes,
Load more button displayed only when we have more notes
image

image


src/presentation/pages/Home.vue
```diff
@@ -1,43 +1,57 @@
 <template>
   <ThreeColsLayout>
```
neSpecc, line 2: add large dimensions, please

src/presentation/pages/Home.vue
```diff
@@ -1,43 +1,57 @@
...
+    <Container>
```
neSpecc, line 3: could you make all the block clickable?

src/presentation/layouts/ThreeColsLayout.vue
```diff
@@ -32,6 +32,7 @@
 .main-content {
   width: var(--layout-content-width);
   padding: 0 var(--spacing-ml);
+  box-sizing: content-box;
```
TatianaFomina, line 35: why `content-box`?

e11sy: for now width of main content layout is fixed and should not be rescaled (without content box main layout would be 700 px, with this box sizing 732 px as it should be)

src/presentation/pages/Home.vue
```diff
@@ -49,13 +80,43 @@ import { useAppState } from '@/application/services/useAppState';
...
-const { noteList, loadMoreNotes } = useNoteList();
+const { noteList, load, loadPreviousPage, currentPage } = useNoteList();
+
+const nextPageNotes = ref<NoteList | null>(null);
+
+/**
+ * On mount of the page we will know if there are more motes to load on next page
+ */
+onMounted(async () => {
+  nextPageNotes.value = await load(currentPage.value + 1);
```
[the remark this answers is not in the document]
e11sy: what if we have 30 notes (full first page), then we will be able to load second, but it would be empty

src/application/i18n/messages/en.json
```diff
@@ -1,6 +1,7 @@
 {
   "appTitle": "NoteX",
-  "loadMore":"Load More",
+  "loadMore": "Load More",
+  "getBack": "Get Back",
```
elizachi, line 4: Whe second word with capital letter?

neSpecc, line 4: is it still used?

src/presentation/pages/Home.vue
```diff
@@ -1,44 +1,68 @@
...
       :level="1"
-      :class="$style['home__heading']"
+      :class="$style['page-header']"
     >
       {{ $t('home.title') }}
     </Heading>
 
-    <div v-if="!user">
+    <div v-if="user === null">
       <p>{{ $t('home.authText') }}</p>
```
neSpecc, line 35: could you include this message to the Row and Container as well? And probably add a Login button to the right
image

neSpecc: and I guess the Load More button should not be displayed in this case

src/presentation/pages/Home.vue
```diff
@@ -1,44 +1,68 @@
...
-      </Button>
+        <Card
+          :title="getTitle(note.content)"
+          :subtitle="getSubtitle(note)"
+          :src="note.cover || undefined"
+          orientation="vertical"
+        />
+      </RouterLink>
     </div>
+    <Button
```
neSpecc, line 59: make it secondary, please

src/presentation/pages/Home.vue
e11sy: should we display create new note block when user is not authorized?

neSpecc: no

neSpecc (approved):

elizachi (approved):

PR #263, opened 2024-07-17
feat(store): session and persistant stores implemented

e11sy:
- implemented session store (it would be updated on page reload)
- implemented persistant store (it would not be updated on page reload), based on window.localStorage
- fixed login required logic (now when user is undefined, service will wait until user is null of actually user)

### DIFF
--- a/src/infrastructure/storage/openedPage.ts
+++ b/src/infrastructure/storage/openedPage.ts
@@ -21,7 +21,6 @@ export class OpenedPagesStore extends PersistantStore<OpenedPagesStoreData> {
    * @param page - page that had beed opened by user
    */
   public addOpenedPage(page: OpenedPage): void {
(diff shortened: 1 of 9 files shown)


src/infrastructure/storage/openedPage.ts
```diff
@@ -11,24 +11,29 @@ export type OpenedPagesStoreData = {
...
-    this.data.openedPages = [];
+    super(['openedPages']);
   }
 
   /**
    * Function for adding record to opened pages storage when user opens new page
    * @param page - page that had beed opened by user
    */
   public addOpenedPage(page: OpenedPage): void {
+    console.log(this.data.openedPages, typeof this.data.openedPages);
```
neSpecc, line 24:
```suggestion
```

elizachi (approved):

neSpecc (approved):

PR #266, opened 2024-07-19
chore(settings): fixed jumping theme changer

e11sy:
## Problem 
Since row height is not fixed, it depends on content,
when we add icon `check` row's height becomes larger for some screens (`text-ui-base-medium` font-size set in rem, so for some screens font-size would be less or than 20px, so when icon added height of one certain row would be changed)

## Solution
Now icon `check` is always displayed, but with opacity 0, for current theme with opacity 1
- with icon `check`
image
- without icon `check`
image

### DIFF
--- a/codex-ui/src/vue/components/row/Row.vue
+++ b/codex-ui/src/vue/components/row/Row.vue
@@ -102,6 +102,7 @@ defineProps<{
   }
 
   &__center {
+    height: 20px;
     flex: 1;
     display: flex;
     flex-direction: column;
--- a/codex-ui/src/vue/components/row/Row.vue
+++ b/codex-ui/src/vue/components/row/Row.vue
@@ -102,7 +102,6 @@ defineProps<{
   }
 
   &__center {
-    min-height: 20px;
     flex: 1;
     display: flex;
     flex-direction: column;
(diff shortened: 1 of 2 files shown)


src/presentation/pages/Settings.vue
```diff
@@ -58,10 +58,12 @@
                 <DarkColorShemeIcon v-if="scheme === 'Dark'" />
               </template>
               <template
-                v-if="colorScheme === scheme.toLowerCase()"
                 #right
               >
-                <Icon name="Check" />
+                <Icon
+                  name="Check"
+                  :style="colorScheme === scheme.toLowerCase() ? 'opacity: 1;' : 'opacity: 0;'"
```
TatianaFomina, line 65: bad practice to set inline styles. Better to use CSS classes

TatianaFomina:
Maybe set constant height instead of always rendered icons?

codex-ui/src/vue/components/row/Row.vue
```diff
@@ -102,6 +102,7 @@ defineProps<{
   }
 
   &__center {
+    min-height: 20px;
```
neSpecc, line 105: don't hardcode it inside a Row component. Row's slot can contain any element, so preventing jumping should be placed in a component where slot is used.

neSpecc (approved):

elizachi (approved):

TatianaFomina (approved):

PR #271, opened 2024-08-18
fix(&nbsp): a bug of header contains the &nbsp fixed

### LINKED ISSUE #257: [ Bug ] `&nbsp;` in tab title when note header contains spaces
Image

dependentmadani:
## Problem:
When the first input in the note contains spaces, the header of the note show `&nbsp;`. 
resolves #257 

## Cause:
The noteContent get stored as HTML entity. For this reason, the title shows the `&nbsp;`, whenever there is occurrence of space in the first block of the note.

## Solution:
- Target the return value of function in `useNote` service and replace all occurrences of `&nbsp;` with space.
- The change was made also on the `NoteList` where the bug occurs too.

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -147,7 +147,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     if (!firstNoteBlock || !Boolean(firstNoteBlock.data.text)) {
       return 'Note';
     } else {
-      return firstNoteBlock.data.text.slice(0, limitCharsForNoteTitle);
+      return firstNoteBlock.data.text.replace(/&nbsp;/g, ' ').slice(0, limitCharsForNoteTitle);
     }
   });
 
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -129,8 +129,6 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
 
   const route = useRoute();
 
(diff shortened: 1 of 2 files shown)


neSpecc:
Please, attach an issue correctly

dependentmadani:
The attachment has been updated.

src/application/services/useNote.ts
```diff
@@ -147,7 +147,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     if (!firstNoteBlock || !Boolean(firstNoteBlock.data.text)) {
```
neSpecc, line 147: it would be better if we will reuse the `getTitle()` util here

dependentmadani: I think there will be an issue of using I18n at the top of a `setup` function.

neSpecc: you can pass a `defaultTitle` as an optional argument, for example

neSpecc (approved):

src/application/services/useNote.ts
```diff
@@ -128,7 +129,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
 
   const route = useRoute();
 
-  const limitCharsForNoteTitle = 50;
+  // const limitCharsForNoteTitle = 50;
```
e11sy, line 132: remove unused comments

neSpecc (approved):

e11sy (approved):

PR #275, opened 2024-11-07
feat: display the note parents structure

### LINKED ISSUE #256: [ Feature ] - display parent note structure in note header
This issue is related to the [api issue](https://github.com/codex-team/notes.api/issues/275)
make sure, that api part is done

## Problem
image
In [disingn system](https://www.figma.com/design/eQr2z1Cx54OIomb0Mn0PzZ/NoteX?node-id=425-4063&t=JgLWyLl0T0sKFf4y-0) we have note parents structure on note page

### Cases
| Structure                                                                                                   | Presentation                                                        |
| ----------------------------------------------------------------------------- | --------------------------------------------------- | 
| `RootParent` > `Actual note`                                                                     | `RootParent` > `Actual note`                               | 
| `RootParent` > `IntermediateNote` > `ActualNote`                                  | `RootParent` > `IntermediateNote` > `ActualNote` |
| `RootParent` > `IntermediateNote1` > `IntermediateNote2` > `ActualNote` | `RootParent` > `...` > `Actual note`                     |

## Solution
- [x] Make method in `useNote` service that would format data got from api into presentation format
- [x] Update `NoteDTO` entity for it to contain new object with note parents info ( optional get rid of parentNote )
``` ts
NoteParents : 
[
  {
    noteId: Note['id'],
    content: Note['content']
  }
]
```
- [ ] Update Note page ( usage of `NoteHeader` component here ) as in [disingn system](https://www.figma.com/design/eQr2z1Cx54OIomb0Mn0PzZ/NoteX?node-id=425-4063&t=JgLWyLl0T0sKFf4y-0)
note that note names should be clickable

dependentmadani:
## Problem:
The issue is that when accessing a note, you would not know the note parent hierarchy, or the note parent of the actual note you are accessing.

## Solution:
As describe in issue #256, we will be displaying the parent note structure at the top left of the note, following the tasks required to insure that the display of the note structure would be efficient.

Resolves #256


src/application/services/useNote.ts
```diff
@@ -158,6 +158,13 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    */
   const parentNote = ref<Note | undefined>(undefined);
 
+  /**
+   * Note parents of the actual note
+   *
+   * Actual note by default
+   */
+  const noteParents = ref<Note[]>([]);
```
e11sy, line 166: i think that making `noteParents` reactive is a bad thing to do
We can update them only by request to api `/GET note` so i can't see the case, when we update noteParents and imidiately change displayed content somewhere

src/application/services/useNote.ts
```diff
@@ -265,6 +273,23 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+      throw new Error('note id is not defined');
+    }
+    let presentationFormat = '';
+
+    for (let value of noteParents.value) {
+      presentationFormat += getTitle(value.content) + ' > ';
+    }
+    presentationFormat += noteTitle.value;
+
+    return presentationFormat;
```
e11sy, line 290: `parentNote` title should be clickable so just merge titles into one string is not a solution. 

Also can't see the case with `RootNoteTitle > ... > CurrentNote`

dependentmadani: I think we can use anchor tag `<a>` can solve the merge of titles into one string.

src/application/services/useNote.ts
```diff
@@ -265,6 +273,23 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     parentNote.value = undefined;
   }
 
+  /**
+   * Format the received note parents into presentation format
```
neSpecc, line 277: Description is not clear

src/application/services/useNote.ts
neSpecc: it's better to do it in a component

src/application/services/useNote.ts
```diff
@@ -70,6 +70,11 @@ interface UseNoteComposableState {
    */
   unlinkParent: () => Promise<void>;
 
+  /**
+   * Get the format of note parents
+   */
+  formatNoteParents: () => Note[];
```
e11sy, line 76: improve description please

e11sy:
This pr does not resolve #256 because presentations part is still not done.
If you want to update note page with new component in a separate pr, then remove `resolves` in pr description

src/application/services/useNote.ts
```diff
@@ -70,6 +70,11 @@ interface UseNoteComposableState {
    */
   unlinkParent: () => Promise<void>;
 
+  /**
+   * Returns an array of Note objects representing the formatted note parents.
+   */
+  formatNoteParents: () => Note[];
```
neSpecc, line 76: from the "composable" patter its better to export "noteParents" object instead

src/application/services/useNote.ts
```diff
@@ -265,6 +278,35 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     parentNote.value = undefined;
   }
 
+  /**
+   * Reform the received note parents from api into presentation format.
+   * @returns An array of Note objects representing the formatted note parents.
+   * @throws {Error} If the note id is not defined.
+   */
```
neSpecc, line 285: I dont really understand what this method is doing. Could you please improve jsdoc to make it more clear? Why API format is not suitable?

src/presentation/pages/Note.vue
```diff
@@ -4,11 +4,18 @@
...
-          note && 'updatedAt' in note && note.updatedAt
-            ? t('note.lastEdit') + ' ' + getTimeFromNow(note.updatedAt)
-            : t('note.lastEdit') + ' ' + 'a few seconds ago'
-        }}
+        <div v-if="noteParents.length">
+          <span
+            v-for="(parent, index) in displayedParents"
+            :key="parent.id"
+          >
+            <a @click="handleParentClick(parent)">{{ newGetTitle(parent.content) }}</a>
```
neSpecc, line 12: use RouterLink instead of `a`

src/presentation/pages/Note.vue
```diff
@@ -4,11 +4,18 @@
       :style="{ '--opacity': id && note ? 1 : 0 }"
     >
       <template #left>
-        {{
-          note && 'updatedAt' in note && note.updatedAt
-            ? t('note.lastEdit') + ' ' + getTimeFromNow(note.updatedAt)
-            : t('note.lastEdit') + ' ' + 'a few seconds ago'
-        }}
+        <div v-if="noteParents.length">
+          <span
```
neSpecc, line 8: span seems redunant

src/presentation/pages/Note.vue
```diff
@@ -103,6 +112,61 @@ function redirectToNoteSettings(): void {
...
+}
+
+/**
+ * Handle parent click
+ *
+ * @param {object} parent - The parent object
+ * @param {string} parent.id - The ID of the parent note
+ */
+function handleParentClick(parent: { id: string }): void {
+  if (parent.id !== 'ellipsis') {
```
neSpecc, line 131: seems like a magic string

src/presentation/pages/Note.vue
```diff
@@ -103,6 +112,61 @@ function redirectToNoteSettings(): void {
...
+  return noteParents;
+});
+
+/**
+ * Update note cover
+ *
+ * @param {OutputData | { title: string }} content - The content of the note
+ * @returns {string} - The title of the note
+ */
+function newGetTitle(content: OutputData | { title: string }): string {
```
neSpecc, line 162: we already have such a method

src/application/services/useNote.ts
```diff
@@ -265,6 +278,35 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+    if (currentId.value === null) {
+      return [];
+    }
+    let presentationFormat: Note[] = [];
+
+    if (noteParents.value.length === 0) {
+      presentationFormat.push({
+        id: currentId.value,
+        content: note.value?.content as NoteContent,
+      });
```
neSpecc, line 296: why you are adding current note as a parent? I don't think that it is a good idea

src/application/services/useNote.ts
```diff
@@ -265,6 +278,35 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     parentNote.value = undefined;
   }
 
+  /**
+   * Reform the array of note parents by adding the actual note id and content.
```
neSpecc, line 282: still don't understand the problem. API returns id and content, why do we need to change them?

src/presentation/pages/Note.vue
```diff
@@ -4,11 +4,17 @@
       :style="{ '--opacity': id && note ? 1 : 0 }"
     >
       <template #left>
-        {{
-          note && 'updatedAt' in note && note.updatedAt
-            ? t('note.lastEdit') + ' ' + getTimeFromNow(note.updatedAt)
-            : t('note.lastEdit') + ' ' + 'a few seconds ago'
-        }}
+        <div>
```
neSpecc, line 7: div seems redundant

src/presentation/pages/Note.vue
```diff
@@ -4,11 +4,17 @@
...
+          <RouterLink
+            v-for="(parent, index) in displayedParents"
+            :key="index"
+            :to="{ path: `/note/${parent.id ? parent.id : noteId}` }"
+            @click.prevent="handleParentClick($event, parent)"
```
neSpecc, line 12: probably you don't need to handle click manually since you are using router link

src/presentation/pages/Note.vue
```diff
@@ -103,6 +110,51 @@ function redirectToNoteSettings(): void {
   router.push(`/note/${props.id}/settings`);
 }
 
+/**
+ * Navigate to a specific note
+ *
+ * @param {number} parentId - The ID of the parent note to navigate to
+ */
+function navigateToParent(parentId: string): void {
```
neSpecc, line 118: seems redundant

src/presentation/pages/Note.vue
```diff
@@ -103,6 +110,51 @@ function redirectToNoteSettings(): void {
...
+function handleParentClick(event: Event, parent: Note): void {
+  if (getTitle(parent.content) !== '...') {
+    navigateToParent(parent.id);
+  } else {
+    event.preventDefault();
+  }
+}
+
+/**
+ * Displayed parents
```
neSpecc, line 138: bad description

src/application/services/useNote.ts
TheGeniusOfEternity: I am not sure how to export it only after it will be filled by load function

src/application/services/useNote.ts
TheGeniusOfEternity: there is some issue that this property isn't empty only after load() for called. So without some check I can't provide it to component because it would be empty and won't be updated. Lets leave this as ref?

src/presentation/components/breadcrumbs/BreadCrumbs.vue
```diff
@@ -0,0 +1,50 @@
...
+import { RouterLink } from 'vue-router';
+import { computed } from 'vue';
+import { Note } from '@/domain/entities/Note.ts';
+import { Icon } from '@codexteam/ui/vue';
+
+const props = defineProps<{
+  noteParents: Note[];
+}>();
+/**
+ * Note's parents hierarchy
```
e11sy, line 28:
```suggestion
 * Note parents hierarchy
```

e11sy (approved):

src/presentation/pages/Note.vue
```diff
@@ -4,11 +4,9 @@
       :style="{ '--opacity': id && note ? 1 : 0 }"
     >
       <template #left>
-        {{
-          note && 'updatedAt' in note && note.updatedAt
-            ? t('note.lastEdit') + ' ' + getTimeFromNow(note.updatedAt)
-            : t('note.lastEdit') + ' ' + 'a few seconds ago'
-        }}
```
neSpecc, line 11: you should not remove "Last edit" info. It should be aligned to the right side:

image

neSpecc (approved):

PR #277, opened 2024-11-16
fix: Redirect to 404 Page When Note is Not Found

### LINKED ISSUE #276: Redirect to 404 Page When Note is Not Found
## Description:
Currently, when a note is not found (i.e., when the API returns a notFound response), the web application continues to display the skeleton loader for the note. This behavior doesn't inform the user that the note couldn't be found.

## Steps:
- Modify the code to detect when the API returns a notFound response (i.e., status code 404).
- When this happens, redirect the user to the /error route (or an equivalent 404 page).
- Ensure that the 404 page properly displays the error message and code.

Tushar504:
- This PR adds try catch block to redirect user to 404 not found error page as described in Issue Closes #276 
- updates the useNote service load method.


src/application/services/useNote.ts
```diff
@@ -166,12 +166,16 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
-    parentNote.value = response.parentNote;
+    try {
+      const response = await noteService.getNoteById(id);
+
+      note.value = response.note;
+      canEdit.value = response.accessRights.canEdit;
+      noteTools.value = response.tools;
+      parentNote.value = response.parentNote;
+    } catch (error) {
+      void router.push('/error');
```
neSpecc, line 177: I'd suggest to separate 404 errors and other errors. If 4xx error happen, show 404 error page. In case of other errors, show 500 "Unexpected error" page

Tushar504: request you to review new changes

src/application/services/useNote.ts
```diff
@@ -166,12 +167,20 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+      note.value = response.note;
+      canEdit.value = response.accessRights.canEdit;
+      noteTools.value = response.tools;
+      parentNote.value = response.parentNote;
+    } catch (error) {
+      if (error instanceof ApiError) {
+        void router.push('/500');
+      } else {
+        void router.push('/400');
+      }
```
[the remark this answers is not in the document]
Tushar504: request you to review new changes

src/application/router/routes.ts
```diff
@@ -181,19 +181,20 @@ const routes: RouteRecordRaw[] = [
...
+      pageTitleI18n: 'pages.error',
       discardTabOnLeave: true,
     },
-    props: {
-      code: 404,
-    },
+    props: route => ({
+      code: route.query.code,
+      customMessage: route.query.message,
+    }),
```
e11sy, line 197: 404 page was actually needed (this was a syntax for each unexpected url, that would show 404 `page not found`)

we need to create separate /error page, that should have `statusCode` property and show different info by different status code

e11sy: also better to use route.params for `code` prop

src/domain/entities/errors/ApiError.ts
```diff
@@ -0,0 +1,16 @@
+import DomainError from './Base';
+
+/**
+ * Domain error thrown when unknown error occurs
+ */
+export default class ApiError extends DomainError {
+  /**
```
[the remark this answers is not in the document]
Tushar504: Actually, currently it is not needed, I had created the ApiError to consider it as default error, to check api error response is instance of ApiError or not,

src/application/services/useNote.ts
```diff
@@ -166,12 +167,20 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+    try {
+      const response = await noteService.getNoteById(id);
+
+      note.value = response.note;
+      canEdit.value = response.accessRights.canEdit;
+      noteTools.value = response.tools;
+      parentNote.value = response.parentNote;
+    } catch (error) {
+      if (error instanceof NotFoundError) {
+        void router.push(`/error/404?message=${error.message}`);
```
neSpecc, line 179: It's better to specify error message directly, instead of using `error.message`. And use i18n for that.

Other code looks good for me

src/application/router/routes.ts
```diff
@@ -195,6 +194,22 @@ const routes: RouteRecordRaw[] = [
...
+    path: '/error/:code',
+    component: ErrorPage,
+    meta: {
+      layout: 'fullpage',
+      pageTitleI18n: 'pages.error',
+      discardTabOnLeave: true,
+    },
+    props: route => ({
+      code: route.params.code,
+      customMessage: route.query.message,
```
neSpecc, line 210: I dont see where `?message` is composed. And it's better to use a prop instead of adding a message to the URL.

Tushar504: you want 'page not found' message right, so on error component it is getting that message from i18n based on error code, if you want custom message then we can pass it as query

neSpecc: it is unused at the moment, so you can remove it.
When adding it in a future, it's better to pass custom message via props instead of URL

src/presentation/pages/Error.vue
```diff
@@ -44,6 +50,17 @@ const message = computed(() => {
 
   return t('errors.default');
 });
+
+useHead({
+  title: t(`errors.${props.code}`),
+});
+
+const openPageInfo = {
+  title: t(`errors.${props.code}`),
```
neSpecc, line 59: I don't see these messages in the dictionary

src/presentation/pages/Error.vue
```diff
@@ -44,6 +50,17 @@ const message = computed(() => {
 
   return t('errors.default');
 });
+
+useHead({
+  title: t(`pages.error`),
+});
+
+const openPageInfo = {
+  title: te(`errors.${props.code}`) ? t(`errors.${props.code}`) : t(`errors.default`),
```
neSpecc, line 59: we don't have this i18n token in dictionary. You should add it or use existing key

src/application/services/useNote.ts
```diff
@@ -166,12 +167,21 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     /**
      * @todo try-catch domain errors
      */
```
neSpecc, line 169:
```suggestion
```

neSpecc (approved):

src/domain/entities/errors/ApiError.ts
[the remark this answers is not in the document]
e11sy: anyway, i think that this is not correct semantically, `DomainError` is more specific than `ApiError`

src/presentation/pages/Error.vue
```diff
@@ -44,6 +50,17 @@ const message = computed(() => {
 
   return t('errors.default');
 });
+
+useHead({
+  title: t(`pages.error`),
+});
+
+const openPageInfo = {
+  title: te(`error.${props.code}`) ? t(`error.${props.code}`) : t(`pages.error`),
```
e11sy, line 59: t(`pages.error`) would be just an `Error` message, that is not actually informative, i think that we should add a default error message like 

`Oops! Something went wrong`

src/domain/entities/errors/ApiError.ts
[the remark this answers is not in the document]
e11sy: hey, @neSpecc what do you think about this structure of errors from api

- `Error`
- `ApiError` extends `Error`
- `DomainError` extends `ApiError` (since we could implement another errors like `StorageError` etc. and that would be also error of api)

then you would be able to check that error is instance of `ApiError` for each `DomainError` and `StorageError`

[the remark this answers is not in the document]
neSpecc: > DomainError extends ApiError

Domain shouldn't know anything about API. It communicated only with Repository. API is what repository uses to retrieve data under the hood.

Based on your code (especially `switch` in `NotesApiTransport`), `ApiError` is just the default DomainError thrown on the same level as NotFoundError and others. 

Ideally, `NotesApiTransport` should throw `NotesApiError`. Repository should handle it somehow and throw `NotesRepositoryError` if needed. Then domain can handle it and throw `DomainError` if needed. 

But for simplicity we decided to throw DomainError from the NotesApiTransport. 

So I think `ApiError extends DomainError` is ok as well as `NotFoundError extends DomainError`.

But I'd suggest to rename ApiError to something more business-logic-specific. Like `ResourceUnavailableError`

src/infrastructure/transport/notes-api/index.ts
```diff
@@ -48,12 +51,15 @@ export default class NotesApiTransport extends AuthorizableTransport {
          */
         switch (status) {
           case 401:
-          case 403:
             return new UnauthorizedError(errorText);
+          case 403:
+            return new ForbiddenError(errorText);
```
neSpecc, line 56: I think, both 401 and 403 should be handled like `UnauthorizedError`.

src/infrastructure/transport/notes-api/index.ts
```diff
@@ -48,12 +51,15 @@ export default class NotesApiTransport extends AuthorizableTransport {
...
           case 404:
             return new NotFoundError(errorText);
+          case 406:
+            return new NotAcceptableError(errorText);
```
neSpecc, line 60: seems redundant, you can handle 406 like other 404+ errors.

src/application/services/useNote.ts
```diff
@@ -163,15 +166,30 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+        case error instanceof UnauthorizedError:
+          void router.push('/error/401');
+          break;
+        case error instanceof ForbiddenError:
+          void router.push('/error/403');
+          break;
+        default:
+          void router.push('/error/500');
+      };
+    }
```
e11sy, line 192: we can make it a bit more straightforward by passing actual `error.name` as prop of `/error` route

then we woun't need this switch case construction

src/domain/entities/errors/Base.ts
```diff
@@ -2,4 +2,14 @@
  * Domain Error — own class describing exceptions in our business-logic.
  * For example, when user is not authorized to perform some action.
  */
-export default class DomainError extends Error {}
+export default class DomainError extends Error {
+  public statusCode?: number;
```
neSpecc, line 6: it should be listed in constructor param, isn't it?

e11sy (approved):

src/domain/entities/errors/NotFound.ts
```diff
@@ -7,8 +7,9 @@ export default class NotFoundError extends DomainError {
   /**
    * Constructor for NotFound error
    * @param message - Error message
+   * @param statusCode - Error status code
    */
-  constructor(message: string = 'NotFound') {
+  constructor(message: string = 'NotFound', public statusCode: number) {
```
neSpecc, line 12: maybe it should be optional? and 404 by default

neSpecc (approved):

e11sy (approved):

Issue #273, opened 2024-10-30
[ Feature ] Codex-ui - implement dropdown component 


e11sy:
## Problem
Implement dropdown component for Codex-UI
Check [this figma file](https://www.figma.com/design/YmJc2Vqmev25xZgXic5bjL/CodeX-Design-System?node-id=1663-1809&node-type=canvas&t=qIGKW0sQSCDbkLhW-0) for design

## Steps
- Create dropdowncomponent inside of the `/codex-ui/src/vue/components` directory 
- Create demonstration file for this component inside of the `/codex-ui/dev/pages` directory

e11sy:
implemented in #279

PR #278, opened 2024-11-28
feat: Dropdown Component for Codex-UI

DeadCreator:
[Feature] Dropdown Component for Codex-UI implementation (#273)
Currently only single select available


codex-ui/src/vue/components/dropdown/Dropdown.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+  <div
+    ref="dropdown"
+    :class="$style['dropdown']"
+  >
+    <div
```
neSpecc, line 6: It's better to use Button component

DeadCreator:
Some problems with test on the first commit, think it's easier to start a new branch

PR #279, opened 2024-11-28
[Feature] Select component for Codex UI

DeadCreator:
Added Dropdown component implementation (#273), but in component only one option is selectable. Also updated icons package version in package json's

### DIFF
--- a/codex-ui/dev/pages/components/Select.vue
+++ b/codex-ui/dev/pages/components/Select.vue
@@ -1,17 +1,24 @@
 <template>
   <PageHeader>
-    Dropdown
+    Select
     <template #description>
       Component of the form that allows you to select one or more options from the list (currently one)
     </template>
   </PageHeader>
-  <Select :items="options" />
+  <Select
+    :items="options"
+    :default-item="defaultItem"
+  />
 </template>
 
 <script setup lang="ts">
 import PageHeader from '../../components/PageHeader.vue';
-import { ContextMenuItem, Select } from '../../../src';
+import { ContextMenuItem, DefaultItem, Select } from '../../../src';
 
+const defaultItem: DefaultItem = {
+  title: 'Choose an option',
+  onActivate: () => {},
+};
 const options: ContextMenuItem[] = [
   {
     type: 'default',
@@ -26,17 +33,11 @@ const options: ContextMenuItem[] = [
     onActivate: console.log,
   },
   {
-    type: 'separator',
-  },
-  {
     type: 'default',
     title: 'Image',
     icon: 'Picture',
     // eslint-disable-next-line no-console
     onActivate: console.log,
-  },
-  {
-    type: 'separator',
   },
   {
     type: 'default',
--- a/codex-ui/dev/pages/components/Select.vue
+++ b/codex-ui/dev/pages/components/Select.vue
@@ -5,14 +5,14 @@
       Component of the form that allows you to select one or more options from the list (currently one)
     </template>
   </PageHeader>
-  <Dropdown :items="dropdownItems" />
+  <Select :items="options" />
 </template>
 
 <script setup lang="ts">
 import PageHeader from '../../components/PageHeader.vue';
-import { ContextMenuItem, Dropdown } from '../../../src';
+import { ContextMenuItem, Select } from '../../../src';
 
-const dropdownItems: ContextMenuItem[] = [
+const options: ContextMenuItem[] = [
   {
     type: 'default',
     title: 'Header',
--- a/codex-ui/dev/pages/components/Select.vue
+++ b/codex-ui/dev/pages/components/Select.vue
@@ -5,10 +5,47 @@
       Component of the form that allows you to select one or more options from the list
     </template>
   </PageHeader>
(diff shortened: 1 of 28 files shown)


codex-ui/src/vue/components/dropdown/Dropdown.vue
```diff
@@ -0,0 +1,81 @@
...
+    <Button
+      :class="$style['dropdown__btn']"
+      :icon="activeItem.icon"
+      trailing-icon="BracketsVertical"
+      secondary
+      @click="togglePopover"
+    >
+      {{ activeItem.title }}
+    </Button>
+    <div :style="showPopover ? {'display' : 'block'} : {'display': 'none'}">
```
neSpecc, line 15: use the Popover component as a wrapper

e11sy:
why file in `dev/pages` is called `Select.vue` if component name is dropdown...

codex-ui/src/vue/components/dropdown/Dropdown.vue
```diff
@@ -0,0 +1,90 @@
...
+  onActivate: () => {},
+};
+const activeItem = ref(defaultValue);
+
+/* Main function to update selected item */
+const updateActiveItem = (item: Item) => {
+  if (item.type === 'default' || !item.type) {
+    activeItem.value = Object.create(item);
+    // eslint-disable-next-line no-console
+    activeItem.value.onActivate = console.log;
```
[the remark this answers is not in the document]
e11sy: also this works wrong

this works wrong:

- If i open popover by pressing dropdown button, then isOpened becomes `true`
- Then i select any option in context menu, popover becomes hidden, and title of the button updates
- If i press popover button one more time it makes isOpened `false` and nothing changes on page (i expect popover to be opened again)

so we should manually `hide()` popover, when new item selected or popover closed by pressing anywhere else

e11sy (changes requested):

codex-ui/dev/pages/components/Select.vue
```diff
@@ -1,14 +1,51 @@
 <template>
   <PageHeader>
-    Select
+    Dropdown
```
e11sy, line 3: now it is select

e11sy:
can't see the icon of the select
image

DeadCreator:
Strange, locally everything is fine...Maybe the reason is that the icon was recently added to icons pack

codex-ui/dev/pages/components/Select.vue
```diff
@@ -1,14 +1,51 @@
...
+  {
+    type: 'default',
+    title: 'Image',
+    icon: 'Picture',
+    // eslint-disable-next-line no-console
+    onActivate: console.log,
+  },
+  {
+    type: 'separator',
+  },
```
neSpecc, line 40: too many separators, lets leave just one

codex-ui/src/vue/components/select/Select.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <Button
+    :icon="activeItem.icon"
+    trailing-icon="BracketsVertical"
+    secondary
+    @click="!isOpen ? togglePopover($event.currentTarget, {vertically: 'below', horizontally: 'left'}) : hide()"
```
neSpecc, line 6: `toggle` means that it includes hide internally. So I would suggest to move `isOpen` check inside it.

codex-ui/src/vue/components/select/Select.vue
```diff
@@ -0,0 +1,63 @@
...
+        items: items,
+      },
+    },
+    align,
+  });
+};
+
+/* Default item value for select on page load */
+const defaultValue: Item = {
+  title: 'Choose an option',
```
neSpecc, line 40: should be customizable via prop

codex-ui/src/vue/components/select/Select.vue
```diff
@@ -0,0 +1,63 @@
...
+  type: 'default',
+  onActivate: () => {},
+};
+const activeItem = ref(defaultValue);
+
+/* Main function to update selected item */
+const updateActiveItem = (item: Item) => {
+  if (item.type === 'default' || !item.type) {
```
neSpecc, line 48: this check is being repeated several times, os it can be utilized

codex-ui/src/vue/components/select/Select.vue
```diff
@@ -0,0 +1,63 @@
...
+    activeItem.value = Object.create(item);
```
neSpecc, line 49: why Object.create is used?

codex-ui/src/vue/components/select/Select.vue
```diff
@@ -0,0 +1,61 @@
...
+  if (!isOpen.value) {
+    showPopover({
+      targetEl: el,
+      with: {
+        component: ContextMenu,
+        props: {
+          items: items,
+        },
+      },
+      align,
```
neSpecc, line 36: lets also pass `width: 'fit-target'` property to make popover width same as  a button width

neSpecc (approved):

e11sy (approved):

PR #280, opened 2024-12-09
fix(ui): popover toggle fix

neSpecc:
Right now you can't call "hide" by click on the popover target element, because [onClickOutside](https://vueuse.org/core/onClickOutside/) will toggle isOpen back.

Now the target element will be ignored in onClickOutside


DeadCreator (approved):
Tried on Select Component - looks like everything is ok 👍

PR #306, opened 2025-02-22
feat(styles): replaced cdn fonts to local

### LINKED ISSUE #304: Imp[styles]: use downloaded version of Inter font
## Problem
Now Inter font is being downloaded from CDN
Image

this significantly slows down page loading

## Solution
Use downloaded version of the font as css file that will be included in the app

DeadCreator:
Added Inter and JetBrainsMono fonts into codex-ui package. This resolves #304


e11sy:
solves does not link pr to the issue, see docs [here](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

e11sy:
also i think that we do not need all of those font customisations (probably bold, semibold and regular will serve our needs)

neSpecc (commented):
I'd suggest moving font to the /src/fonts instead of src/styles

e11sy (approved):

PR #307, opened 2025-02-24
Fix : Note content overrides when switching between notes

### LINKED ISSUE #303: [Fix]: Note content overrides when switching between notes
# Problem
On Notes page, Note content overrides when we switch while editing the current note.

#Solution
Need to work on Props and State Management of Note Page.

Tushar504:
This pr updates the useNote service and Note vuejs component to Resolve #303


e11sy (commented):
remove commented lines if they are not needed

e11sy (approved):

PR #308, opened 2025-02-26
feat: Imp[NoteSettings] add func to sort team in correct order

3388ShubinPavel:
- Add sortMembers function that sorts members in order Creator -> Writer -> Reader
- Change view to display team from sortedMember array

### DIFF
--- a/src/presentation/components/team/Team.vue
+++ b/src/presentation/components/team/Team.vue
@@ -57,8 +57,7 @@ const props = defineProps<{
  * Sorts the team members based on their role and whether they are the creator of the note.
  *
  * @param members - Array of TeamMember objects to be sorted.
- // eslint-disable-next-line jsdoc/require-returns-type
- * @returns Sorted array of TeamMember objects.
+ * @returns TeamMember - Sorted array of TeamMember objects.
  */
 const sortMembers = (members: TeamMember[]): TeamMember[] => {
   return members.sort((a, b) => {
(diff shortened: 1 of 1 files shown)


src/presentation/components/team/Team.vue
```diff
@@ -33,24 +33,62 @@
...
 
-defineProps<{
+const { t } = useI18n();
+
+const props = defineProps<{
   /**
    * Team of the current note
    */
-  team: Team;
+  team: TeamMember[];
```
neSpecc, line 49: unexpected change

neSpecc (approved):

src/presentation/components/team/Team.vue
```diff
@@ -50,7 +53,42 @@ defineProps<{
...
+    if (isCreatorA && !isCreatorB) {
+      return -1;
+    }
+    if (!isCreatorA && isCreatorB) {
+      return 1;
+    }
+
+    if (!roleOrder[a.role] && !roleOrder[b.role]) {
+      return 0;
+    }
```
e11sy, line 81: this condition will work if both roles are write and you will get `(!0 && !0)` condition

however below you have condition that will return same thing for same data `(0 === 0)`

so this condition seems redundant

e11sy (changes requested):

e11sy (approved):

neSpecc (approved):

PR #311, opened 2025-03-08
Fix | added check on noteTitle's url patch

### LINKED ISSUE #309: [Bug]: Note settings tab patches to note name
Sometimes we can see, that note settings tab patches just to note title 

Image

Need to find a bug, probably in openedPages storage

TheGeniusOfEternity:
Added check to prevent Note Settings page patch. This resolves #309


src/application/services/useNote.ts
```diff
@@ -322,12 +322,16 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+    const routes = ['/settings', '/history', '/error'];
+
+    if (routes.every(rt => !route.path.includes(rt))) {
+      patchOpenedPageByUrl(
+        route.path,
+        {
+          title: currentNoteTitle,
+          url: route.path,
+        });
+    }
```
e11sy, line 334: maybe we can just check for regex match?
it should match the `/note\/*{10}/` where 10 is length of the noteId

then we know that route is related to the note

src/application/services/useNote.ts
```diff
@@ -322,12 +322,14 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
   });
 
   watch(noteTitle, (currentNoteTitle) => {
-    patchOpenedPageByUrl(
-      route.path,
-      {
-        title: currentNoteTitle,
-        url: route.path,
-      });
+    if (!route.path.match('/note/.{10}/')) {
```
e11sy, line 325: now it seems very strange, we patch all the titles that are not related to the note actually?

TheGeniusOfEternity: Yeah, like before changes. Now we just don't patch url if route path contains only note's url

e11sy (approved):

src/application/services/useNote.ts
```diff
@@ -379,13 +379,14 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
 
   watch(noteTitle, (currentNoteTitle) => {
-    patchOpenedPageByUrl(
-      route.path,
-      {
-        title: currentNoteTitle,
-        url: route.path,
-      });
-
+    if (!route.path.match('/note/.{10}/')) {
```
neSpecc, line 382: why not use `route.name`?

TheGeniusOfEternity: Hmm, yeah, it would be better, will replace!

neSpecc (approved):

PR #321, opened 2025-04-09
fix(marketplace): prevent layout shrink when right column is missing

Dobrunia:
### Problem
On the NoteX Web if the right or left column is not rendered, the main layout shrinks unexpectedly.
Снимок экрана 2025-04-09 в 03 20 24

### Solution
Added a placeholder sidebar element with fixed width when the right or left column slot is empty. This ensures the layout remains consistent even if no content is provided for the right or left column.
Снимок экрана 2025-04-09 в 03 21 00

### DIFF
--- a/@codexteam/ui/package.json
+++ b/@codexteam/ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codexteam/ui",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "type": "module",
   "scripts": {
     "dev": "vite",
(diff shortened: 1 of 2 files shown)


@codexteam/ui/src/vue/layout/page-block/PageBlock.vue
```diff
@@ -30,6 +34,10 @@
     >
       <slot name="right" />
     </div>
+    <div
```
neSpecc, line 37: it can be simplified by moving v-if to a `slot`

neSpecc (approved):

@codexteam/ui/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@codexteam/ui",
-  "version": "0.1.0",
```
e11sy, line 3: i think that we need to add publish ci

Dobrunia: I'll make this a separate feature

e11sy (approved):

PR #324, opened 2025-05-14
Fix: Display current note in BreadCrumbs component

### LINKED ISSUE #323: [Fix]: display current note in BreadCrumbs component
In BreadCrumbs component, besides note's hierarchy, we need to display the note itself

TheGeniusOfEternity:
By this PR, in note's breadcrumbs also its title will be displayed

This resolves #323

### DIFF
--- a/src/presentation/components/breadcrumbs/BreadCrumbs.vue
+++ b/src/presentation/components/breadcrumbs/BreadCrumbs.vue
@@ -2,7 +2,7 @@
   <RouterLink
     v-for="(parent, index) in displayedParents"
     :key="index"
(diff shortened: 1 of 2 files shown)


src/application/services/useNote.ts
```diff
@@ -263,6 +260,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
        */
       const noteCreated = await noteService.createNote(content, specifiedNoteTools, parentId);
 
+      noteParents.value = (await noteService.getNoteById(noteCreated.id)).parents;
```
neSpecc, line 263: add comment explaining this statement, please

PR #326, opened 2025-07-15
Note hierarchy appreance

devnull-1337:
Note hierarchy appreance changes.
User can now toggle if he wants to show left menu of note hierarchy.

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -1,4 +1,4 @@
-import { ref, type Ref, type MaybeRefOrGetter, computed, toValue, watch } from 'vue';
+import { onMounted, ref, type Ref, type MaybeRefOrGetter, computed, toValue, watch } from 'vue';
 import { noteService, editorToolsService } from '@/domain';
 import type { Note, NoteContent, NoteId } from '@/domain/entities/Note';
 import type { NoteTool } from '@/domain/entities/Note';
@@ -9,13 +9,6 @@ import DomainError from '@/domain/entities/errors/Base';
 import useNavbar from './useNavbar';
 import { getTitle } from '@/infrastructure/utils/note';
 import type { NoteHierarchy } from '@/domain/entities/NoteHierarchy';
-
-/**
- * Note-specific load options
- */
-export interface NoteLoadOptions {
-  loadHierarchy?: boolean;
-}
 
 /**
  * Creates base structure for the empty note:
@@ -105,11 +98,16 @@ interface UseNoteComposableState {
   noteHierarchy: Ref<NoteHierarchy | null>;
 
   /**
-   * Load note by id with configurable options
+   * Load note by id
    * @param id - Note identifier
-   * @param loadOptions - Load options for customizing what data to fetch
    */
-  load: (id: NoteId, loadOptions?: NoteLoadOptions) => Promise<void>;
+  load: (id: NoteId) => Promise<void>;
+
+  /**
+   * Load note hierarchy separately when needed
+   * @param id - Note identifier
+   */
+  loadHierarchy: (id: NoteId) => Promise<void>;
 }
 
 interface UseNoteComposableOptions {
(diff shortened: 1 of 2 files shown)


src/application/services/useNote.ts
```diff
@@ -96,6 +98,18 @@ interface UseNoteComposableState {
...
+   * Note settings composable
+   */
+  noteSettings: Ref<NoteSettings | null>;
+  /**
+   * Updates the cover image of the note.
+   * @param id - The identifier of the note whose cover should be updated.
+   * @param data - The new cover image as binary data (Blob).
+   * @returns A Promise that resolves when the cover has been updated.
+   */
+  updateCover: (id: string, data: Blob) => Promise<void>;
```
e11sy, line 112: why did you move this method to `useNote`?

src/presentation/pages/Note.vue
```diff
@@ -99,7 +99,7 @@ const props = defineProps<{
 
 const noteId = toRef(props, 'id');
 
-const { note, noteTools, save, noteTitle, canEdit, noteParents, noteHierarchy } = useNote({
+const { note, noteTools, save, noteTitle, canEdit, noteParents, noteHierarchy, noteSettings, updateCover } = useNote({
```
e11sy, line 102: unexpected change

src/application/services/useNote.ts
devnull-1337: It was part of the noteSettings which was in notePage but now its moved to useNote so moved it here

src/application/services/useNote.ts
```diff
@@ -320,15 +341,6 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
 
-  onMounted(() => {
-    /**
-     * If we have id, load note and note hierarchy
-     */
-    if (currentId.value !== null) {
-      void load(currentId.value);
-    }
-  });
-
```
e11sy, line 331: why load removed?

src/presentation/pages/Note.vue
```diff
@@ -203,18 +197,33 @@ function transformNoteHierarchy(noteHierarchyObj: NoteHierarchy | null, currentN
...
       useHead({
         title: t('note.new'),
       });
+    } else {
+      // Load note settings first
+      await loadNoteSettings(props.id);
+
+      // Create load options based on note settings
+      const loadOptions: NoteLoadOptions = {
+        loadHierarchy: noteSettings.value?.showNoteHierarchy === true,
```
e11sy, line 222: you can just cast to boolean 
```suggestion
        loadHierarchy: !!noteSettings.value?.showNoteHierarchy,
```

src/application/services/useNote.ts
```diff
@@ -206,7 +221,13 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
       noteTools.value = response.tools;
       parentNote.value = response.parentNote;
       noteParents.value = response.parents;
-      void getNoteHierarchy(id);
+
+      // Default to false if not specified
+      const shouldLoadHierarchy = loadOptions?.loadHierarchy ?? false;
```
e11sy, line 226:
```suggestion
      const shouldLoadHierarchy = !!loadOptions?.loadHierarchy;
```

src/application/services/useNote.ts
```diff
@@ -206,7 +221,13 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+
+      if (shouldLoadHierarchy) {
+        void getNoteHierarchy(id);
+      }
```
e11sy, line 230: this should be a separate composable method that allows to get NoteHierarchy when it is needed

leave onMounted note load as it was

src/presentation/pages/Note.vue
devnull-1337: not required removed

src/application/services/useNote.ts
devnull-1337: as of now kept it in the same composable and got onmounted back

neSpecc (approved):

PR #336, opened 2026-01-12
fix: switching between notes may override note

redeck1:
Causes: watch() updated the tab by route.path.
When switching between notes quickly, load() could complete after switching to another note, but route.path would point to the new note.
State race during loading/saving:
Async load() and save() did not check that the currentId had not changed during the request. PATCH /note/:id updated the local state for a different note.

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -383,7 +383,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
 
     // Recursively update child notes
     if (hierarchy.childNotes) {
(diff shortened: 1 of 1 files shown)


src/application/services/useNote.ts
```diff
@@ -255,9 +265,14 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
      */
     const specifiedNoteTools = resolveToolsByContent(content);
 
+    /**
+     * Id may be changed
```
neSpecc: JSdoc is not clear for me, please explain the case.

PR #341, opened 2026-02-01
Fix/codexui

quangtuanitmo18:
Description: 
- the button states are broken: Secondary, Disabled now look like Primary
- Some of our controls require a specific theme, for example, <Button destructive> takes the red color from the Red theme. Therefore, the Red theme should be enabled by default.

image

### DIFF
--- a/@codexteam/ui/package.json
+++ b/@codexteam/ui/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@codexteam/ui",
-  "version": "0.2.0-rc.5",
+  "version": "0.2.1",
   "type": "module",
   "sideEffects": [
     "*.css",


@codexteam/ui/src/styles/index.pcss
```diff
@@ -2,3 +2,7 @@
 @import './typography.pcss';
 @import './mixins.pcss';
 @import './z-axis.pcss';
+/**
+ * Export theme red by default (always included in base styles)
```
neSpecc:
```suggestion
 * Import 'red' theme by default since it is used for desctructive-action styles
```

neSpecc (approved):

Issue #350, opened 2026-05-06
Popover component does not reposition on scroll or window resize


quangtuanitmo18:
Currently, when the Popover component is opened, it calculates its absolute position relative to the target element perfectly. However, if the user scrolls the page (or a scrollable container) or resizes the browser window while the popover is active, the popover remains fixed in its initial static coordinates and detaches from its target element.

Additionally, there is a minor bug in the horizontal position calculation where window.scrollX is missing, which could cause offset issues if the page is scrolled horizontally.

<img width="961" height="312" alt="Image" src="https://github.com/user-attachments/assets/142c9e6d-5747-479e-947f-0a0ddb840658" />

<img width="638" height="336" alt="Image" src="https://github.com/user-attachments/assets/0a67932a-7d28-4cad-8e6d-abbab9b43bdc" />

PR #351, opened 2026-05-06
Popover component does not reposition on scroll or window resize

quangtuanitmo18:
Related to #350


Dobrunia (approved):

neSpecc (approved):

Reversean:
### Code review

Found 1 issue:

1. `updatePosition` checks `targetElement.value` for null but not whether the element is still attached to the DOM. If the trigger element is removed while the popover is open (e.g. a team member row is deleted via `MoreActions.vue`), `getBoundingClientRect()` returns all zeros and the next scroll moves the popover to the document's top-left corner. Adding `!targetElement.value.isConnected` to the guard would fix this.

https://github.com/codex-team/notes.web/blob/f0f6d78b334c5c27a015842f7db792df9372d48c/@codexteam/ui/src/vue/components/popover/usePopover.ts#L130-L136

<sub>If this review was useful, react with 👍. Otherwise, 👎.</sub>

PR #355, opened 2026-07-08
Fix: ensure tools are configured before editor renders on note switch

7eliassen:
# Problem:
image
Occasionally, when switching between notes, some editor.js blocks may not be displayed correctly.

# Root Cause:
image
*toolsUserConfigLoaded and isEditorReady already true when the function is just called*

This issue occurs because the toolsUserConfigLoaded and isEditorReady flags are not properly reset. As a result, the editor may render before the tools are fully configured, causing certain blocks to fail to display.

# What done:
- Added logic to reset toolsUserConfigLoaded and isEditorReady before calling loadToolsScripts, ensuring that the editor waits for the tools to be properly configured before rendering.
- Destroy the editor reliably on unmount to prevent bug (screenshot bellow) when the previous editor instance wasn't cleaned up.
image

- Added a load counter to discard stale async results and prevent race conditions during rapid note switching.

### DIFF
--- a/src/application/services/useNoteEditor.ts
+++ b/src/application/services/useNoteEditor.ts
@@ -18,12 +18,6 @@ interface UseNoteEditorOptions {
   noteContentResolver: () => NoteContent | undefined;
 
   /**
-   * Function to check if the note is a draft
-   * In draft we wont wait for note tools loading
-   */
-  isDraftResolver: () => boolean;
-
-  /**
    * Flag indicating that user can edit the note
    */
   canEdit: Ref<boolean>;
@@ -85,18 +79,21 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
 
   /**
    * Combine note and user tools
-   * Undefined when user or note is not loaded
+   * Returns undefined when note tools are not loaded yet to prevent
+   * premature editor rendering with an incomplete tools set
    */
   const noteAndUserTools = computed<EditorTool[] | undefined>(() => {
-    const noteTools = toValue(options.noteTools) || [];
-    const userTools = toValue(userEditorTools) ?? [];
+    const noteTools = toValue(options.noteTools);
 
     /**
-     * If tools are not loaded yet, return undefined
+     * If note tools are not loaded yet, return undefined to prevent
+     * premature editor rendering
      */
     if (noteTools === undefined) {
       return undefined;
     }
+
+    const userTools = toValue(userEditorTools) ?? [];
 
     /**
      * Return unique array of tools grouped by tool.name
--- a/src/application/services/useNoteEditor.ts
+++ b/src/application/services/useNoteEditor.ts
@@ -144,6 +144,9 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
       return;
     }
 
+    isEditorReady.value = false;
+    toolsUserConfigLoaded.value = false;
+
     await loadToolsScripts(tools);
   }, {
     immediate: true, // load tools if they are passed to the composable immediately
--- a/src/application/services/useNoteEditor.ts
+++ b/src/application/services/useNoteEditor.ts
@@ -77,6 +77,13 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
   const toolsUserConfigLoaded = ref<boolean>(false);
 
   /**
+   * Incremented on each new load request to discard stale async results.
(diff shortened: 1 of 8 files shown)


src/application/services/useNoteEditor.ts
```diff
@@ -144,7 +147,28 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
       return;
     }
 
-    await loadToolsScripts(tools);
+    const loadId = ++currentLoadId;
+
+    isEditorReady.value = false;
+    toolsUserConfigLoaded.value = false;
+
+    const loadedConfig = await loadToolsScripts(tools);
```
Reversean: `loadToolsScripts` can throw. Since `isEditorReady` is now reset to `false` *before* this `await`, a failed download leaves the note stuck with no editor rendered at all.

Might be worth wrapping this in a try/catch and surfacing an error on failure? Or maybe restoring some usable state?

src/application/services/useNoteEditor.ts
```diff
@@ -144,7 +147,28 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
...
+
+    /**
+     * If a newer load request has superseded this one — discard stale results
+     * to prevent overwriting state with tools from a previous note.
+     */
+    if (loadId !== currentLoadId) {
```
Reversean: Could you add test case so the race condition this is meant to fix doesn't regress silently?

src/application/services/useNoteEditor.ts
```diff
@@ -144,7 +147,28 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
       return;
     }
 
-    await loadToolsScripts(tools);
+    const loadId = ++currentLoadId;
+
+    isEditorReady.value = false;
```
Reversean, line 158: `isEditorReady` now resets on every `noteAndUserTools` recompute, not just on note switch — e.g. installing a tool from the marketplace live-updates the shared `userEditorTools` ref while a note is open. Worth checking that case: does `<Editor>` visibly flash/remount instead of just picking up the new tool like before?

src/application/services/useNoteEditor.ts
```diff
@@ -102,10 +109,12 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
...
 
   /**
-   * Downloads passed tools scripts and toggles-on the isEditorReady flag
+   * Downloads passed tools scripts and returns the loaded config object.
+   * Does not mutate shared state — the caller is responsible for applying the result
    * @param toolsConfigs - tools to download
+   * @returns loaded tools config
    */
-  async function loadToolsScripts(toolsConfigs: EditorTool[]): Promise<void> {
+  async function loadToolsScripts(toolsConfigs: EditorTool[]): Promise<Record<string, { class: EditorjsConfigTool; inlineToolbar: boolean }>> {
```
Reversean: Might be worth extracting a shared type alias so `loadToolsScripts` and `toolsUserConfig` can't drift apart?

Reversean (changes requested):

src/application/services/useNoteEditor.ts
7eliassen: I added a handler to set the value of  `isEditorReady.value` to true anyway. So, even if an error occurs, the note will be displayed with the default tools

src/application/services/useNoteEditor.ts
7eliassen: Added `EditorjsToolsConfig` type and updated return type annotation for `loadToolsScripts`

src/application/services/useNoteEditor.ts
Reversean: Since there is not test infrastructure, set it up in #359.

Covered with tests in #360.

Reversean (approved):

7eliassen:
Some changes have been made due to a discovered bug.

A bug was discovered related to double rendering of the editor when switching to another note:

When `NoteId` changed, `noteAndUserTools` was recalculated due to the `isDraftResolver()` call. This triggered an editor render with the old content. Shortly after, `noteAndUserTools` was recalculated again due to changes in `options.noteTools`, and the correct note was displayed.

**Fix:**
- Removed the `isDraftResolver()` method
- Drafts now define `noteTools` as an empty array
- Added an authentication check to properly handle the `userTools` value

neSpecc (approved):

PR #356, opened 2026-07-08
Refactor: optimize notes covers rendering

7eliassen:
### Problem
The homepage currently suffers from poor loading performance. The note list is not displayed until all cover images have finished downloading, causing significant delays — especially on slower connections.

### Solution
- **Parallelized cover requests** — switched to asynchronous requests for downloading covers, allowing multiple connections to be established simultaneously for faster overall load time.
- **Decoupled cover loading from metadata fetching** — note metadata now loads first, while covers are fetched independently in the background.
- **Reactive cover rendering** — covers appear on cards progressively as soon as each individual image finishes downloading, rather than waiting for the entire batch to complete.

### Preview
Peek 2026-07-08 13-125

### Implementation Details

**`noteList.service.ts`**
- `getNoteList()` now returns metadata only (cover downloads are excluded)
- Added new `loadCover()` method for per-note asynchronous cover fetching

**`useNoteList.ts`**
- Introduced `loadCovers()` which downloads all covers in parallel using `Promise.all()` and updates each note reactively upon completion
- `loadNextPage()` triggers `loadCovers()` in the background after the list is rendered, ensuring a smooth user experience

### Performance Improvements
Tested on Linux in Firefox and Chrome

Home page loading (30 notes with covers):
| Browser | Before | After |
|---------|--------|-------|
| Firefox | ~20s   | ~5s   |
| Chrome  | ~12s   | ~4s   |

### DIFF
--- a/src/application/services/useNoteList.ts
+++ b/src/application/services/useNoteList.ts
@@ -34,6 +34,11 @@ interface UseNoteListComposableState {
    * Loading state
    */
   isLoading: Ref<boolean>;
+
+  /**
+   * Set of note IDs whose cover downloads are currently in-flight
+   */
+  coversLoading: Ref<Set<string>>;
 }
 
 /**
@@ -86,7 +91,7 @@ export default function (onlyCreatedByUser = false): UseNoteListComposableState 
    * Prevents duplicate requests when loadMoreNotes() fires again before
    * covers for the previous page have finished loading.
    */
-  const coversLoading = new Set<string>();
+  const coversLoading = ref(new Set<string>());
 
   /**
    * Load cover images for all notes in the list in the background
@@ -118,11 +123,11 @@ export default function (onlyCreatedByUser = false): UseNoteListComposableState 
       /**
        * If this note's cover is already being fetched, skip it
        */
-      if (coversLoading.has(item.id)) {
+      if (coversLoading.value.has(item.id)) {
         return;
       }
 
-      coversLoading.add(item.id);
+      coversLoading.value = new Set(coversLoading.value).add(item.id);
 
       try {
         const url = await noteListService.loadCover(item.id, item.cover);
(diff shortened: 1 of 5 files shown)


neSpecc (commented):
can we show cover skeleton while image is still loading?

src/application/services/useNoteList.ts
```diff
@@ -69,17 +69,62 @@ export default function (onlyCreatedByUser = false): UseNoteListComposableState
...
+        /**
+         * Update the specific note's cover reactively so the card renders the image
+         */
+        list.items[index] = {
+          ...currentItem,
+          cover: url,
+        };
+      }
+    }));
   };
```
Reversean, line 156: `loadCovers()` re-scans the whole `noteList.value.items` array on every call (not just the page that was just loaded). If `loadMoreNotes()` fires again before an earlier page's covers finish downloading (easy to hit on the slow connections), the still-pending items from the previous page get requested a second time in parallel with the first request. 

Might be worth tracking in-flight note ids or only passing the newly-appended items into `loadCovers()`, so fast second page load can't re-trigger downloads for covers that are already underway?

Reversean (changes requested):

src/application/services/useNoteList.ts
7eliassen: Added a `coversLoading` set that contains currently downloading covers to prevent re-downloading.

7eliassen:
Peek 2026-07-21 20-57

Added a loading skeleton to note card covers when images are being fetched.

### Changes

__`Card.vue`__ — Added an `isCoverLoading` prop. When `true`, the cover area shows a shimmer animation instead of the image.

__`useNoteList.ts`__ — Made `coversLoading` reactive (previously a plain `Set`) so the template can observe which covers are still loading.

__`NoteList.vue`__ — Passes the new `isCoverLoading` prop to each card based on whether its cover is currently being fetched.

neSpecc (approved):

Reversean (approved):

PR #357, opened 2026-07-09
feat: user cannot delete himself from team

7eliassen:
Prevent a user from removing themselves from the note's team member list.

## Changes

- The "Remove" context menu item is now only shown for team members whose email does not match the current user's email

- The popover button is only shown when there is at least one menu item to display

### DIFF
--- a/src/presentation/components/team/MoreActions.vue
+++ b/src/presentation/components/team/MoreActions.vue
@@ -18,9 +18,9 @@ import { type TeamMember } from '@/domain/entities/Team';
 import { useI18n } from 'vue-i18n';
 import { NoteId } from '@/domain/entities/Note';
 import useNoteSettings from '@/application/services/useNoteSettings';
+import { useAppState } from '@/application/services/useAppState';
 
 const { removeMemberByUserId } = useNoteSettings();
-
 const props = defineProps<{
   /**
    * Team member data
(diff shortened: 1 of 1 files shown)


src/presentation/components/team/MoreActions.vue
```diff
@@ -32,28 +32,31 @@ const props = defineProps<{
...
 const { showPopover, hide } = usePopover();
 const { confirm } = useConfirm();
 
 const triggerButton = ref<HTMLButtonElement>();
 
-const menuItems: ContextMenuItem[] = [
-  {
+const menuItems: ContextMenuItem[] = [];
+
+if (props.teamMember.user.email !== user.value?.email) {
```
neSpecc: make computed property instead, thus we can handle user value change

7eliassen: Done. Check last commit

src/presentation/components/team/MoreActions.vue
```diff
@@ -32,34 +32,41 @@ const props = defineProps<{
...
-    onActivate: async () => {
-      hide();
-      await handleRemove(props.teamMember);
-    },
-  },
-];
+const menuItems = computed<ContextMenuItem[]>(() => {
+  const items: ContextMenuItem[] = [];
+
+  if (props.teamMember.user.email !== user.value?.email) {
```
Reversean: Comparing by `email` – since both `user.value` and `teamMember.user` are `User` objects with an `id`, and `handleRemove` below already uses `member.user.id`, might be better to key this check off `user.id` too?

Matches how the API guard in codex-team/notes.api#304 identifies the user, and avoids any edge case where email casing differs between the session user and the team record.

Reversean (changes requested):

src/presentation/components/team/MoreActions.vue
7eliassen: Changed email to id

neSpecc (approved):

Reversean (approved):

PR #362, opened 2026-08-05
Fix: overwrite note content on switching

### LINKED ISSUE #363: Bug: Note content may be overwritten
## Summary

When switching between notes, it is possible to overwrite the contents of another note if the save() method was called immediately before switching to the other note.

## Steps to reproduce:

- Create 2 notes with different content
- Modify the content of one note, which triggers the save() method
- Quickly switch to the other note

## Presentation

https://github.com/user-attachments/assets/38098c2a-832b-45df-bb01-bd7fe711a89e

7eliassen:
## Problem

When switching between notes, the previously opened note's content could be applied to the newly opened note. Two root causes:

1. The previous note's editor stayed mounted while the new note loaded, so its handlers remained alive and could write the old content into the new note.
2. An in-flight `save()` used the id of the currently open note at completion time, so it wrote the old content into whatever note was open then.

## Solution

**1-st Fix: destroy the old editor on note change.** `useNoteEditor` now watches `noteId` and resets `isEditorReady` on every change. Since the `<Editor>` is rendered with `v-if="isEditorReady"`, the previous editor is unmounted/destroyed and a fresh one is mounted only after the new note's data and tools are ready. A live old editor can no longer save into the new note.

**2-nd Fix: bind each save to the note it was started for.** `noteChanged()` captures the note id at save start and passes it to `save()`, so an in-flight save always targets the correct note. Cached content (`lastUpdateContent`) and the note cover are only updated when the current note still matches the saved note, preventing stale data from leaking between notes.

## Key changes

- **`useNote.ts`** — `save()` now takes the captured note id and uses it for the update; cached content is only stored when the current note hasn't changed. Removed the now-unneeded `isNoteSaving` flag and the related draft-save skip logic.
- **`useNoteEditor.ts`** — added a `noteId` option and reset the editor state when the note changes.
- **`Note.vue`** — captures the note id at save time and passes it through; the cover is only updated when the captured id still matches the current note.

### DIFF
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -96,6 +96,13 @@ interface UseNoteComposableState {
    * Note hierarchy
    */
   noteHierarchy: Ref<NoteHierarchy | null>;
+
+  /**
+   * Returns the id of the note created by the most recent save() on a new note
+   * Used to distinguish "same note just got an id after save" from
+   * "switched to a different existing note"
+   */
+  getLastCreatedNoteId: () => NoteId | null;
 }
 
 interface UseNoteComposableOptions {
@@ -183,6 +190,12 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    * null by default
    */
   const noteHierarchy = ref<NoteHierarchy | null>(null);
+
+  /**
+   * Id of the note created by the most recent save() on a new note
+   * Used to skip the reload after save so the editor doesn't get recreated
+   */
+  let lastCreatedNoteId: NoteId | null = null;
 
   /**
    * get note hierarchy
@@ -273,6 +286,19 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
        * @todo try-catch domain errors
        */
       const noteCreated = await noteService.createNote(content, specifiedNoteTools, parentId);
+
+      /**
+       * Remember the created note id so the editor can avoid
+       * recreating itself when the route switches from "new note" to the newly created note id
+       */
+      lastCreatedNoteId = noteCreated.id;
+
+      /**
+       * Store the saved content so the navbar title reflects it
+       */
+      if (currentId.value === currentNoteId) {
+        lastUpdateContent.value = content;
+      }
 
       /**
        * Replace the current route with note id
@@ -389,6 +415,14 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
       return;
     }
 
+    /**
+     * If the note was just created via save() and is still a draft (no id yet),
+     * skip the reload to avoid recreating the editor with the same content.
+     */
+    if (newId === lastCreatedNoteId && note.value !== null && !('id' in note.value)) {
+      return;
+    }
+
     void load(newId);
   });
 
@@ -416,5 +450,6 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     noteParents,
     parentNote,
     noteHierarchy,
+    getLastCreatedNoteId: () => lastCreatedNoteId,
   };
 }
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -202,6 +202,7 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
       const response = await noteService.getNoteById(id);
 
       note.value = response.note;
+      lastUpdateContent.value = response.note.content;
       canEdit.value = response.accessRights.canEdit;
       noteTools.value = response.tools;
       parentNote.value = response.parentNote;
@@ -290,9 +291,12 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
     }
 
     /**
-     * Store just saved content in memory
+     * Store just saved content in memory only if the current note hasn't changed
+     * This prevents race conditions when switching between notes quickly
      */
-    lastUpdateContent.value = content;
+    if (currentId.value === currentNoteId) {
+      lastUpdateContent.value = content;
+    }
 
     isNoteSaving.value = false;
   }
--- a/src/application/services/useNote.ts
+++ b/src/application/services/useNote.ts
@@ -55,7 +55,7 @@ interface UseNoteComposableState {
   /**
    * Creates/updates the note
    */
-  save: (content: NoteContent, parentId: NoteId | undefined) => Promise<void>;
+  save: (content: NoteContent, parentId: NoteId | undefined, currentNoteId: NoteId | null) => Promise<void>;
 
   /**
    * Returns list of tools used in note
@@ -244,8 +244,9 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
    * Saves the note
    * @param content - Note content (Editor.js data)
    * @param parentId - Id of the parent note. If null, then it's a root note
+   * @param currentNoteId - Id of the current note
    */
-  async function save(content: NoteContent, parentId: NoteId | undefined): Promise<void> {
+  async function save(content: NoteContent, parentId: NoteId | undefined, currentNoteId: NoteId | null): Promise<void> {
     if (note.value === null) {
       throw new Error('Note is not loaded yet');
     }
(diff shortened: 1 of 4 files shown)


src/application/services/useNoteEditor.ts
```diff
@@ -83,6 +97,27 @@ export const useNoteEditor = function useNoteEditor(options: UseNoteEditorOption
...
+    () => toValue(options.noteId),
+    (newId, oldId) => {
+      /**
+       * Same note just got an id after save — keep the editor as-is
+       */
+      if (oldId === null && newId !== null && options.getLastCreatedNoteId !== undefined && newId === options.getLastCreatedNoteId()) {
+        return;
+      }
+
+      isEditorReady.value = false;
```
Reversean: **Critical bug**: This watcher leaves `isEditorReady` set to `false` permanently when switching from an existing note to a new note (e.g. clicking "+").

**Why this happens**:
1. When switching from an existing note to a new note, `noteId` changes from `'some-id'` to `null`.
2. This watcher sets `isEditorReady.value = false;`.
3. In `useNote.ts`, `resetNote()` does **not** reset `noteTools.value`. As a result, `noteAndUserTools` does not change, and its watcher (`watch(noteAndUserTools)`) never fires.
4. Because `isEditorReady.value = true` is only set in `watch(noteAndUserTools)`, `isEditorReady` stays `false` forever. The editor never mounts when creating a new note.

Additionally, `options.noteId` changing does not bump `currentLoadId` in `useNoteEditor`. If tool loading for note A was in flight when switching to note B, note A's tools resolution will still execute and mark `isEditorReady = true` using note A's tools.

src/presentation/pages/Note.vue
```diff
@@ -169,8 +177,8 @@ async function noteChanged(data: NoteContent): Promise<void> {
         paddingTop: '100px',
       });
     }
-    if (updatedNoteCover !== null && props.id !== null) {
-      await updateCover(props.id, updatedNoteCover);
+    if (updatedNoteCover !== null && noteIdAtCallTime !== null && noteIdAtCallTime === props.id) {
```
Reversean: **Bug**: Cover screenshots are never saved for newly created notes.

For a new note, `noteIdAtCallTime` is captured before `save()` as `props.id`, which is `null`.
After `save()` resolves and the route updates to the newly created note ID, this check is evaluated:
```ts
if (updatedNoteCover !== null && noteIdAtCallTime !== null && noteIdAtCallTime === props.id) {
```
Because `noteIdAtCallTime` is `null`, `noteIdAtCallTime !== null` evaluates to `false`. The screenshot is taken via `makeElementScreenshot`, but never passed to `updateCover()`. Every newly created note ends up without a cover.

**Recommendation**: Have `save()` return the ID of the saved note (`savedNoteId`), and check `props.id === savedNoteId` before calling `updateCover(savedNoteId, updatedNoteCover)`.

src/application/services/useNote.ts
```diff
@@ -198,10 +212,21 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
+    const loadId = ++currentLoadId;
+
     try {
       const response = await noteService.getNoteById(id);
 
+      /**
+       * If a newer load request has superseded this one — discard stale results
+       * to prevent mismatched content/tools state when switching notes quickly
+       */
+      if (loadId !== currentLoadId) {
```
Reversean: **Race condition in error handling**: The `loadId !== currentLoadId` guard protects against stale responses in the `try` block, but the `catch` block (lines 235-242) has no check.

If note A is requested and fails (e.g. 404, 403, or network abort) after the user has already navigated to note B, the `catch` block executes with `route.path` pointing to note B:
```ts
deleteOpenedPageByUrl(route.path);
if (error instanceof DomainError) {
  void router.push(`/error/${error.statusCode}`);
} else {
  void router.push('/error/500');
}
```
This inadvertently closes note B's tab and redirects to the error page even though note B may be valid or already loaded.

**Recommendation**: Add `if (loadId !== currentLoadId) return;` inside the `catch` block.

src/application/services/useNote.ts
```diff
@@ -255,14 +281,25 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
      */
     const specifiedNoteTools = resolveToolsByContent(content);
 
-    isNoteSaving.value = true;
-
-    if (currentId.value === null) {
+    if (currentNoteId === null) {
```
Reversean: **Race condition / UX bug**: `router.replace` executes unconditionally when draft creation resolves.

If the user starts typing in a new note and then navigates to an existing note while `noteService.createNote` is in flight, `router.replace` will still be invoked when `createNote` finishes:
```ts
await router.replace({
  name: 'note',
  params: {
    id: noteCreated.id,
  },
});
```
This forcibly navigates the user back to the newly created note, interrupting their navigation.

**Recommendation**: Ensure that the user is still on the draft (`currentId.value === null`) before replacing the route.

src/application/services/useNote.ts
```diff
@@ -377,13 +415,11 @@ export default function (options: UseNoteComposableOptions): UseNoteComposableSt
...
-    const isDraftSaving = prevId === null && isNoteSaving.value;
-
     /**
-     * Case for newly created note,
-     * we don't need to re-load it
+     * If the note was just created via save() and is still a draft (no id yet),
+     * skip the reload to avoid recreating the editor with the same content.
      */
-    if (isDraftSaving) {
+    if (newId === lastCreatedNoteId && note.value !== null && !('id' in note.value)) {
```
Reversean: **Architectural / state consistency issue**: 
1. When a new note is created in `save()`, `note.value` is never updated with `noteCreated`. It remains a `NoteDraft` (without `id`, `createdAt`, `updatedAt`, etc.). Because this check skips `load(newId)` specifically when `!('id' in note.value)`, `note.value` remains incomplete until an external reload occurs.
2. `lastCreatedNoteId` is never reset. If the user subsequently navigates to another note and then returns to this note (or opens a draft again), `lastCreatedNoteId` retains stale state and can suppress reloading or editor reset incorrectly.

**Recommendation**: Set `note.value = noteCreated` in `save()`, and properly manage the lifecycle of `lastCreatedNoteId` rather than relying on `!('id' in note.value)` as a sentinel.

src/presentation/pages/HistoryVersion.vue
```diff
@@ -83,6 +83,7 @@ const { noteTitle, save } = useNote({
 const canEdit = ref(false);
 
 const { isEditorReady, editorConfig } = useNoteEditor({
+  noteId,
```
Reversean: **Bug when browsing history**: When switching between different history versions of the same note (`historyId` changes, but `noteId` stays the same), `watch(options.noteId)` in `useNoteEditor` does not fire.

Since `isEditorReady.value = false` was removed from `watch(noteAndUserTools)`, the editor is never unmounted or re-initialized with the new version's content and tool configuration.

**Recommendation**: Pass `historyId` or a version identifier to `useNoteEditor` (or keep editor unmounting tied to tool/content updates for non-draft views) so history version switches properly reset the editor.

src/application/services/useNoteEditor.ts
Reversean: When switching from an existing note to a new note, `isEditorReady` is set to `false`, but `watch(noteAndUserTools)` never fires because `resetNote()` doesn't touch `noteTools`, leaving the editor permanently unmounted.

src/presentation/pages/Note.vue
```diff
@@ -169,8 +177,8 @@ async function noteChanged(data: NoteContent): Promise<void> {
...
+      await updateCover(noteIdAtCallTime as NoteId, updatedNoteCover);
```
Reversean: For new notes, `noteIdAtCallTime` is `null`, so `noteIdAtCallTime !== null` is always false and covers are never saved.

```suggestion
    if (updatedNoteCover !== null && props.id !== null && (noteIdAtCallTime === null || noteIdAtCallTime === props.id)) {
      await updateCover(props.id as NoteId, updatedNoteCover);
```

src/application/services/useNote.ts
Reversean: If an older load request fails after switching notes, the un-guarded `catch` block still deletes the active tab and redirects the user to an error page.

src/application/services/useNote.ts
Reversean: If the user navigates away while a new note is being created, `router.replace` still executes unconditionally and pulls the user away from their active page.

Reversean (approved):

